PR #7, opened 2016-03-30
feature/choropleth legend

GabeMedrash:
This branch breaks up what had been called "range-slider" into its subcomponents, and renames the composed component for what it really is: IHME's choropleth legend. 

In a future PR, this component could be extended to take in any arbitrary SlideHandle component.

It also flattens the directories such that components now live directly within `src` dirs, not nested in `src` -> `components`.


brettvitaz:
Nice increase in coverage! Looks good to me.

PR #11, opened 2016-04-19
change to dataGenerator

aaronKeel:
Made some changes to improve dataGenerator for axisChart.

### DIFF
--- a/src/test-utils/data.js
+++ b/src/test-utils/data.js
@@ -46,17 +46,17 @@ export const dataGenerator = (config = {}) => {
   const yearProducer = {
     initYear: startYear,
     currYear: startYear - length,
-    next: function() {
+    next() {
       this.currYear++;
       return this.currYear;
     }
   };
 
   const trend = {
-    increasing: (p) => {return p + linearGen();},
-    decreasing: (p) => {return p - linearGen();},
-    exponentialGrowth: (p) => {return p * expGen();},
-    exponentialDecay: (p) => {return p * 1 / expGen();}
+    increasing(p) { return p + linearGen(); },
+    decreasing(p) { return p - linearGen(); },
+    exponentialGrowth(p) { return p * expGen(); },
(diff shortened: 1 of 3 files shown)


src/test-utils/data.js
```diff
@@ -1,73 +1,99 @@
...
 
       switch (dataQuality) {
         case 'best':
-          val = randNum;
+          val = newVal;
           break;
         case 'mixed':
-          useInt = randomGenerator.bool(0.75); // 75% chance of true
-          val = useInt ? randNum : null;
+          useNum = random.randomUniform()() < 0.75 ? true : false; // 75% chance of true
```
brettvitaz: do you need to add the ternary operation here?

src/test-utils/data.js
```diff
@@ -1,73 +1,99 @@
...
+    intitialValue * changeFactor,
+    initialDeviation * changeFactor
+  );
+  const expChange = Math.pow(1 / changeFactor, 1 / length);
+  const expGen = random.randomNormal(expChange, 0.01);
 
-  const yearProducer = (function* yearProducer() {
-    for (let year = startYear - length + 1; year <= startYear; year++) {
-      yield year;
+  const yearProducer = {
```
brettvitaz: Why reinvent a generator? If you wanted to get rid of the for loop, you could just use a while (true)

```
    const yearProducer = (function* yearProducer() {
        let currYear = startYear - length;
        while (true) {
            yield ++currYear;
        };
    }());
```

GabeMedrash: I think it's probably OK to ditch generators for now to avoid having to import babel-polyfill--removes a little bit of upfront complexity in getting demos running.

On another note, though, we should keep our "producers" consistently implemented. Given that the valueProducer is needlessly complex, I think it can be refactored to be more consistent with yearProducer. But that can be saved for a future PR.

PR #13, opened 2016-05-10
Feature/slider component

brettvitaz:
This component renders an input slider with the following api:

```
  height: PropTypes.number,
  width: PropTypes.number,
  minValue: PropTypes.number.isRequired,
  maxValue: PropTypes.number.isRequired,
  value: PropTypes.oneOfType([
    PropTypes.number,
    PropTypes.shape({
      min: PropTypes.number,
      max: PropTypes.number
    })
  ]).isRequired,
  labelFunc: PropTypes.func,
  onChange: PropTypes.func.isRequired
```


brettvitaz:
And the track is clickable too. 
Functions as follows:
If one handle is rendered, that handle value will be updated on click.
If two handles is rendered, the handle that is closest to the click will be updated on click.

GabeMedrash:
Well done, @brettvitaz! Just a couple minor, inline comments.

src/ui/slider/src/handle.jsx
```diff
@@ -0,0 +1,90 @@
...
+  componentWillUnmount() {
+    this._interactable.unset();
+  }
+
+  render() {
+    return (
+      <div
+        className={ classNames(style.flag, style[this.props.direction]) }
+        style={ { left: `${this.props.position}px` } }
+        ref="handle"
```
GabeMedrash: we should use the callbacks to handle refs instead of strings; i believe this method is deprecated

src/ui/slider/src/style.css
```diff
@@ -0,0 +1,85 @@
...
+  font-size: 8pt;
+  text-align: center;
+  vertical-align: middle;
+  white-space: nowrap;
+  color: white;
+  background-color: #27ae60;
+  transition: left 0.3s ease-out;
+}
+
+.flag.left {
```
GabeMedrash: for this and similar declarations, i believe we can just do:
```
.left {
  composes: flag
 ...
}
```

src/ui/slider/src/track.jsx
```diff
@@ -0,0 +1,44 @@
...
+      });
+  }
+
+  componentWillUnmount() {
+    this._interactable.unset();
+  }
+
+  render() {
+    return (
+      <div ref="track" className={ style.track }></div>
```
GabeMedrash: same comment with use of ref callback

src/ui/slider/src/handle.jsx
```diff
@@ -0,0 +1,90 @@
...
+        snap: {
+          targets: [getSnapTargetFunc(this.props.snapTarget, {
+            range: Infinity,
+            offset: { x: this.offset }
+          })]
+        }
+      })
+      .styleCursor(false)
+      .on('dragmove', this.props.onMove(this.props.name, -this.offset))
+      .on('dragend', this.props.onEnd(this.props.name));
```
GabeMedrash: I think we need to call a (probably throttled version) of `props.onEnd` on `dragmove`. I'm just imagining someone not letting go of the mouse and wanting to see how data changes as they move through the years.

Additionally, we'll also need to support the use of the arrow keys. Is this something interact can do for us?

src/ui/slider/src/style.css
brettvitaz: I am not convinced that this is the result I am trying to achieve. When using `composes`, the selector is duplicated in the classname. Lets discuss this.

src/ui/slider/src/handle.jsx
brettvitaz: Currently, the `onClick` callback is called from the interact `'dragend'` event, not `'dragmove'`. If the user wants to see updates while dragging this will need to be refactored.

I'm not sure about key press support in interact. I will look into this.

brettvitaz:
Added track filling. 
In the process, I refactored the element that the handles are attached to, and supporting scale/snapTarget logic. 
The Slider now handles new width, min/maxValue props properly.

src/ui/slider/demo/app.jsx
```diff
@@ -0,0 +1,52 @@
...
+      }
+    };
+
+    this.onChange = this.onChange.bind(this);
+    this.nextState = this.nextState.bind(this);
+
+    // setTimeout(this.nextState({ ...this.state, minValue: 1995, maxValue: 2040 }), 2000)
+  }
+
+  nextState(state) {
```
GabeMedrash: this still used?

src/ui/slider/src/slider.jsx
```diff
@@ -1,190 +1,226 @@
...
-      && typeof edgePosition === 'number' && edgePosition === edgePosition) {
-      const diff = containerWidth - edgePosition;
-      return (diff < 0) ? 0 : diff;
+  componentWillReceiveProps(newProps) {
+    // If the extents or width changes, the scale and snapTarget must be recalculated.
+    if ((this.props.minValue !== newProps.minValue) ||
+      (this.props.maxValue !== newProps.maxValue) ||
+      (this.props.width !== newProps.width)) {
+      this.state.scale.domain([newProps.minValue, newProps.maxValue]);
+      this.forceUpdate(this.receiveTrackWidth);
```
GabeMedrash: I think this forceUpdate is unnecessary. If we're already in `componentWillReceiveProps`, then we're already in the update/render cycle

src/ui/slider/src/style.css
```diff
@@ -0,0 +1,82 @@
...
+  border-right-width: 1px;
+  transform: translate(-100%, 0px);
+}
+
+.flag.right {
+  border-left-width: 1px;
+  transform: translate(0px, 0px);
+}
+
+.flag.left.connected {
```
GabeMedrash: Probably worth getting rid of if not used

src/ui/slider/demo/app.jsx
brettvitaz: this is for testing extent and width changes. it might make for a more interesting demo :)

src/ui/slider/src/slider.jsx
brettvitaz: I did not figure out how to make this work without `forceUpdate`. snapTarget needs the actual new width, which it does not get until after `<Track/>` is rendered, which it does not have at this point, hence the post `forceUpdate` callback to `this.receiveTrackWidth`. 

Please let me know if you have a better idea.

GabeMedrash:
I just added one more commit to show the code to produce the slider in the demo, and also to show both a range slider and a single value slider.

On another note, any chance you know why codecov isn't running on this PR?

src/ui/slider/src/slider.jsx
GabeMedrash: I see. Well, `componentDidUpdate` might be the place to do it, but we should investigate if there are any framework-level benefits of making the call to `receiveTrackWidth` there as opposed to this `forceUpdate` in `componentWillReceiveProps`. One pro would be we'd be using the lifecycle methods literately; one con would be we'd end up reproducing the logic you already have in `componentWillReceiveProps` in another lifecycle method. 

As a side note, I updated the demo with a button to change the minValue/maxValue props passed to the sliders so that this more observable.

brettvitaz:
I added `step` to the slider and immediately realized that floating point math issues were causing some calculated values to not be precise. I added a `getFloatPrecision` function to determine the intended float precision based on the step size, and a `valueWithPrecision` function to return a normal looking value. I did not find a solution I liked that doesn't involve converting the number to a string and back again. Let me know if you know of a better way to do this.

I also realized that adding support for a list as an input source is unnecessary because it is easy to use the slider as it is to index into a list. Look at the third demo (just added) to see how I propose to do this.

brettvitaz:
I added support for selecting a handle and handling left and right cursor keys. I'm glad I did this after adding support for step sizes. The current behavior is that the handle must be clicked to be selected (not dragged). Alternatively, it can be automatically selected when dragged. Let me know your thoughts. This works on all 4 major browsers (took a while to figure out how to support Safari and FF)!

GabeMedrash:
👍

PR #14, opened 2016-05-12
Data generator

aaronKeel:
Some improvements were made to the data generator. It now produces unique composite primary keys so you can have different locations, age groups, etc. It can also have multiple value fields that can have upper and lower bound uncertainties. Length is a parameter that can give a number of years in a time series, but the number of actual rows of data could be a lot more than length depending on what the primary keys are. 

I included a description in README.md.

All tests pass. I had to go through the tests that use dataGenerator an adjust for the new implementation.


examples/app.js
```diff
@@ -66,7 +66,7 @@ class App extends React.Component {
         yScaleType="linear"
         extraClasses={['foo', 'bar']}
       >
-        <XAxis style={axisStyle} />
+        <XAxis A />
```
GabeMedrash: lets double check this

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
...
+  //   [{k_1:v_1}, {k_1:v_2}, {k_1:v_3}],
+  //   [{k_2:v_1}, {k_2:v_2}, {k_2:v_3}],
+  //   [{k_3:v_1}, {k_3:v_2}]
+  // ]
+  const keyStore = _.map(primaryKeys, (k) => {
+    return _.map(k.values, (v) => {
+      const obj = {};
+      obj[k.name] = v;
+      return obj;
+    });
```
GabeMedrash: we can make the return of the inner `map` more concise with computed properties, e.g.:
```js
return {
  [k.name]: v
};
```

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
...
+  let uniqKeys = [{}];
+  _.forEach(keyStore, (keyValues) => {
+    const sto = [];
+    _.forEach(keyValues, (keyObj) => {
+      _.forEach(uniqKeys, (rowObj) => {
+        sto.push(Object.assign({}, rowObj, keyObj));
+      });
+    });
+    uniqKeys = sto;
+  });
```
GabeMedrash: this is a good use case for a `reduce` instead of the outermost `forEach`; it will make it a little more explicit that we care about returning an accumulated array as opposed to iterating over an array for some other side effect. for that matter, you may also consider turning the inner `forEach`s into `map`s.

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
...
+
 
-  const trend = {
-    increasing(p) { return p + linearGen(); },
-    decreasing(p) { return p - linearGen(); },
-    exponentialGrowth(p) { return p * expGen(); },
-    exponentialDecay(p) { return p * 1 / expGen(); }
-  };
+  function floor(number) {
```
GabeMedrash: it would be helpful if this had some basic inline documentation about what this does, e.g.:
```js
/**
 * modified floor to take into account...
 * @param number {Number}
 * @return {Number} 
 */
```

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
...
+    return Math.floor(number * 10) / 10;
+  }
 
-  let prev = intitialValue;
+  function amp(range) {
```
GabeMedrash: same comment re documentation

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
...
+  }
 
-  const valueProducer = (() => {
-    return () => {
-      const newVal = trend[dataTrend](prev);
-      prev = newVal;
-      let val;
-      let useNum;
+  function sAxis(range) {
+    return (range[1] + range[0]) / 2;
```
GabeMedrash: same comment re documentation

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
...
+      const rowObj = {};
+      for (let k = 0; k < valueKeys.length; k++) {
+        const valObj = valueKeys[k];
+        const obj = {};
+        // sinusoidal data generator y=Asin(2x/L+B)+D
+        const value = floor(amp(valObj.range) * sin(2 * j / length + k + i) + sAxis(valObj.range));
+        obj[valObj.name] = value;
+        if (valObj.uncertainty) {
+          obj[`${valObj.name}_ub`] = floor(value + amp(valObj.range) / 4);
+          obj[`${valObj.name}_lb`] = floor(value - amp(valObj.range) / 4);
```
GabeMedrash: `amp(valObj.range)` is called three times, this is a good opportunity to cache it in a variable

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
...
+        }
+        Object.assign(rowObj, obj);
```
GabeMedrash: best to use lodash::assign for now ([Object.assign compat](http://kangax.github.io/compat-table/es6/#test-Object_static_methods_Object.assign))

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
...
-    };
+
+  // Populate rows.
+  // [
+  //   {k_1:v_1, k_2:v_1, k_3:v_1, k_a:v_1, k_b:v_1, k_c:v_1, k_d:v_1},
+  //   {k_1:v_2, k_2:v_1, k_3:v_1, k_a:v_2, k_b:v_2, k_c:v_2, k_d:v_2},
+  //   ...
+  // ]
+  const rows = [];
+  _.forEach(valueData, (valArr, i) => {
```
GabeMedrash: might consider a `flatMap` here as well in addition to an inner `map`

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
...
+    const yObj = { year_id: year + i };
+    _.forEach(valArr, (valRow, j) => {
+      const rowObj = { id: cuid() };
+      Object.assign(rowObj, valRow, uniqKeys[j], yObj);
```
GabeMedrash: prefer again lodash::assign

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
-import { map } from 'lodash';
+import _ from 'lodash';
```
GabeMedrash: our style so far has been to pull in only the lodash functions we need per module. e.g.:
```js
import { flatMap, map, forEach, ... } from 'lodash';
```

This way using a [tree-shaking module bundler](http://www.2ality.com/2015/12/webpack-tree-shaking.html), we can limit the size of the bundle

src/test-utils/data.js
```diff
@@ -1,100 +1,144 @@
...
+  for (let j = 0; j < length; j++) {
+    const segment = [];
+    for (let i = 0; i < uniqKeys.length; i++) {
+      const rowObj = {};
+      for (let k = 0; k < valueKeys.length; k++) {
+        const valObj = valueKeys[k];
+        const obj = {};
+        // sinusoidal data generator y=Asin(2x/L+B)+D
+        const value = floor(amp(valObj.range) * sin(2 * j / length + k + i) + sAxis(valObj.range));
+        obj[valObj.name] = value;
```
GabeMedrash: another opportunity for a more concise assignment with computed properties:
```js
 const obj = {
  [valObj.name]: floor(...)
};
```

package.json
```diff
@@ -32,6 +32,7 @@
     "coverage-token": "cat ./coverage/coverage.json | ./node_modules/codecov.io/bin/codecov.io.js",
     "demo": "BABEL_ENV=webpack scripts/bundle-demos.sh",
     "test": "npm run lint && BABEL_DISABLE_CACHE=1 BABEL_ENV=cli $(npm bin)/babel-node $(npm bin)/_mocha --require ./scripts/setup-jsdom.js `./scripts/get-test-directories.sh`",
+    "test:data": "npm run lint && BABEL_DISABLE_CACHE=1 BABEL_ENV=cli $(npm bin)/babel-node $(npm bin)/_mocha --require ./scripts/setup-jsdom.js src/test-utils/test/data.test.js",
```
GabeMedrash: i think we should get rid of this

README.md
```diff
@@ -77,25 +80,50 @@ Property | Required | Type(s) | Description
...
+`valueKeys` is an array of objects that have a `name` property that is a string, a `range` property that is an array of two numbers, and an `uncertainty` property that is a boolean. Value keys are iterated so that their values are within the range specified. If `uncertainty` is true, data generator will produce additional keys of the form `(name)_ub` and `(name)_lb` to represent upper and lower bound uncertainties.
 
-- `import RangeSlider from 'beaut/ui/range-slider';`
-- `import { RangeSlider } from 'beaut/ui';`
+`year` is a number that represents a starting year for a series of years iterated by length. The ouput key is `year_id`.
 
-#### <a name="utility-functions" href="#utility-functions">#</a> Utility Functions
-- `import utils from 'beaut/util';`
-- `import { domain, scale } from 'beaut/utils';`
+`length` is a number for which each unique composite key gets a new value key.
```
GabeMedrash: can you clarify this?

PR #19, opened 2016-05-23
legend

GabeMedrash:
a configurable legend that renders a list of legend items within a <ul>


src/ui/legend/src/legend-item.jsx
```diff
@@ -0,0 +1,125 @@
...
+    PropTypes.string,
+
+    /* or a function to resolve the symbol type, passed the current item */
+    PropTypes.func
+  ])
+};
+
+const defaultProps = {
+  item: {},
+  itemHeight: 23,
```
brettvitaz: maybe instead of a specific style property, take a style object.

src/ui/legend/src/legend-item.jsx
```diff
@@ -0,0 +1,125 @@
...
+    onClear,
+    onClick,
+    renderClear,
+    symbolColorKey,
+    symbolTypeKey
+  } = props;
+  const color = propResolver(item, symbolColorKey);
+  const type = propResolver(item, symbolTypeKey);
+
+  function wrappedOnClick(event) {
```
brettvitaz: combine this with the next function. extract and make a utility funciton

src/ui/legend/src/legend.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+  /* any inline styles to apply to legend wrapper */
+  wrapperStyle: PropTypes.object
+};
+
+const defaultProps = {
+  items: [],
+  itemRenderer: LegendItem,
+  itemHeight: 23,
+  renderClear: true,
```
brettvitaz: possible bug: this breaks the ability to define `renderClear` as:
```<Legend items={...} onClick={...} ... renderClear />```
because the value is true by default.
It must be defined as:
```<Legend items={...} onClick={...} ... renderClear={false} />```

File: src/ui/legend/src/legend.jsx [hunk 4]
```diff
@@ -0,0 +1,136 @@
...
+export default class Legend extends React.Component {
+
+  renderTitle() {
+    const { title, titleRenderer } = this.props;
+    if (!title) return null;
+    return titleRenderer({ title });
+  }
+
+  renderItemList() {
+    // easier to omit the props we don't need than to declare the ones we do
```
brettvitaz: since there are 15 props, this is almost 50/50. i would tend to lean towards being explicit.

src/ui/legend/src/legend.jsx
```diff
@@ -0,0 +1,136 @@
...
+      'title',
+      'titleRenderer',
+      'ulClassName',
+      'wrapperClassName',
+      'wrapperStyle'
+    ]);
+
+    if (!isArray(items) || !items.length) return null;
+    return items.map((item) => {
+      return itemRenderer({ item, ...itemProps });
```
brettvitaz: With regard to behavior, this has the effect of blending the LegendItem children right in line with the Legend (in the react tab of the console), and hides implementation. Not sure if that is a problem, just a note.
Relates to our discussion of checking for a react component.

PR #26, opened 2016-06-07
Scatter plot

aaronKeel:
Two new .jsx files were created: 'multi-scatter' and  'scatter'. This follows the 'multi-line' and 'line' components for multiple datasets or a single dataset. There are tests for each component and a demo of different uses of scatter and multi-scatter. Let me know what you think or if changes need to be made or clarified.


src/ui/shape/index.js
```diff
@@ -2,4 +2,6 @@ export { default as Area } from './src/area';
 export { default as Line } from './src/line';
 export { default as MultiLine } from './src/multi-line';
 export { default as ScatterPlot } from './src/scatter-plot';
```
GabeMedrash: if this supersedes the prior effort ("ScatterPlot"), please delete any code/files that should no longer be used

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+  /*
+    key names for accessing x and y data.
+  */
+  dataAccessors: PropTypes.shape({
+    x: PropTypes.string,
+    y: PropTypes.string
+  }).isRequired,
+
+  clickHandler: PropTypes.func,
```
GabeMedrash: while we have yet to go through the remainder of the repo to update all modules, we've decided it's best to name this prop "onClick"

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+  hoverHandler: PropTypes.func
```
GabeMedrash: re:above, this should be renamed "onHover"

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+const defaultProps = {
+  clickHandler: noop,
+  hoverHandler: noop,
+  dataField: 'values',
+  symbolField: 'type',
+  colorScale: () => { return 'steelblue'; }
+};
+
+const MultiScatter = (props) => {
```
GabeMedrash: per more recent [airbnb style recommendations](https://github.com/airbnb/javascript/tree/master/react#class-vs-reactcreateclass-vs-stateless), please change this to a normal function:
```
export default function MultiScatter(props) {...}
```

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+  /*
+    key name for value of symbol
+  */
+  symbolField: PropTypes.string,
+
+  /*
+    function to transform symbol value to a shape
+  */
+  symbolScale: PropTypes.func,
```
GabeMedrash: if we don't require props, we should provide a default.

src/ui/shape/src/scatter.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+  /*
+    key names for accessing x and y data.
+  */
+  dataAccessors: PropTypes.shape({
+    x: PropTypes.string,
+    y: PropTypes.string
+  }).isRequired,
+
+  clickHandler: PropTypes.func,
```
GabeMedrash: re note above, change to "onClick"

src/ui/shape/src/scatter.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+  hoverHandler: PropTypes.func
```
GabeMedrash: re note above, change to "onHover"

src/ui/shape/src/scatter.jsx
```diff
@@ -0,0 +1,105 @@
...
+};
+const defaultProps = {
+  clickHandler: noop,
+  hoverHandler: noop,
+  symbolType: 'circle',
+  color: 'steelblue'
+};
+
+const Scatter = (props) => {
```
GabeMedrash:
```
export default function Scatter(props) {...}
```

GabeMedrash:
Looks good Aaron, thanks! I left a couple inline comments.

Some general stylistic notes that you have no way of knowing because Brett and I more or less just talked about them and should really be written down somewhere:
1. We think it's probably best to name event handling props like `on${Event}`. E.g., "onClick", "onHover", etc.
2. Also, we should start passing through the actual event object to user-supplied event handling functions. We have a [simple utility function](https://github.com/ihmeuw/ihme-ui/blob/master/src/utils/events.js) to help with that.
3. We still need to update other modules, but we want to transition all of our arrow-function functional components to just regular functions, e.g.:
```
const Foo = (props) => {...};
export default Foo;
```
becomes
```
export default function Foo(props) {...}
```

PR #28, opened 2016-06-14
Add `geoJSONKeyField`

brettvitaz:
to specify the key field on the geoJSON, if it isn't the default 'id'.


GabeMedrash:
thumbs up

PR #29, opened 2016-06-15
Prevent legend item svg resize.

davschne:


GabeMedrash:
Looks good, David--thanks!

PR #33, opened 2016-06-24
Axis label enhancements

brettvitaz:
Notes:
* label is added to `<Axis />` as a prop
* label is placed based on orientation, drawn against the inside edge of the padding
* for label, must supply padding to `<Axis />`
* dimensions prop flattened to width, height
* default orientation removed for `<Axis />`
* default scales are provided for `<[XY]?Axis />` and `<Multi(Line|Scatter) />` to avoid run time warnings
* unit test coverage has been improved

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,7 +8,8 @@ module.exports = {
   "rules": {
     "comma-dangle": 0,
     "no-underscore-dangle": [2, { "allowAfterThis": true }],
-    "arrow-body-style": [2, "always"]
+    "arrow-body-style": [2, "always"],
+    "max-len": [2, 100, 2, { "ignoreUrls": true, "ignoreComments": true }]
   },
   "env": {
     "mocha": true
--- a/src/ui/axis-chart/src/axis-chart.jsx
+++ b/src/ui/axis-chart/src/axis-chart.jsx
@@ -58,28 +58,20 @@ export function calcDimensions(width, height, margins) {
     height: height - (margins.top + margins.bottom)
   };
 }
+
 export default class AxisChart extends React.Component {
   componentWillMount() {
     this.componentWillReceiveProps(this.props);
   }
 
   componentWillReceiveProps(props) {
-    const {
-      width,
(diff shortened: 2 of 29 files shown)


.eslintrc.js
```diff
@@ -8,12 +8,13 @@ module.exports = {
   "rules": {
     "comma-dangle": 0,
     "no-underscore-dangle": [2, { "allowAfterThis": true }],
-    "arrow-body-style": [2, "always"]
+    "arrow-body-style": [2, "always"],
+    "max-len": [2, 100, 2, { "ignoreUrls": true, "ignoreComments": true }]
```
GabeMedrash: nice. yeah, this is annoying to deal with.

src/ui/axis-chart/src/axis-chart.jsx
```diff
@@ -1,132 +1,128 @@
...
+        x: getScale(props.xScaleType)().domain(props.xDomain).range([0, chartDimensions.width]),
+        y: getScale(props.yScaleType)().domain(props.yDomain).range([chartDimensions.height, 0]),
+      },
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const state = {};
+
+    if (propsChanged(this.props, nextProps, ['width', 'height', 'padding'])) {
```
GabeMedrash: i like this a lot

src/utils/props.js
```diff
@@ -0,0 +1,52 @@
...
+
+export const CommonPropTypes = {
+  className: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.array,
+    PropTypes.object,
+  ])
+};
+
+export function oneOfProp(propTypes) {
```
GabeMedrash: do you think it would be worth changing the name of this function to `exactlyOneOfProp` to help disambiguate its purpose from `atLeastOneOfProp`?

brettvitaz: Good idea, I will change to `exactlyOneOfProp`.

PR #34, opened 2016-06-25
Slider enhancements

brettvitaz:
Optimize child component rendering. @GabeMedrash, you were right about how components render, lets discuss.
Use `dX` to move handle instead of `pageX`, which makes handles track cursor at point of click instead of center.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "ihme-react-select": "git+https://github.com/ihmeuw/ihme-react-select.git",
     "interact.js": "^1.2.6",
     "lodash": "^4.5.0",
-    "react": "^15.0.1",
+    "react": "^15.1.0",
     "react-addons-pure-render-mixin": "^15.1.0",
     "react-addons-shallow-compare": "^15.1.0",
     "react-dom": "^15.1.0",
--- a/src/ui/slider/src/slider.jsx
+++ b/src/ui/slider/src/slider.jsx
@@ -223,13 +223,13 @@ export default class Slider extends PureComponent {
   }
 
   render() {
-    const { fontSize, width, wrapperStyles, wrapperClassName, ticks } = this.props;
+    const { fontSize, width, wrapperStyle, wrapperClassName, ticks } = this.props;
     const { snapTarget } = this.state;
 
     const styles = {
       fontSize: `${fontSize}`,
       width: `${width}px`,
-      ...wrapperStyles,
+      ...wrapperStyle,
     };
 
     return (
@@ -266,11 +266,9 @@ Slider.propTypes = {
   /* step between slider values. */
   step: PropTypes.number,
 
-  /* inline styles applied to outermost wrappper */
-  wrapperStyles: PropTypes.object,
-
-  /* any additional classes to add to outermost wrapper */
+  /* class name styles applied to outermost wrapper */
   wrapperClassName: CommonPropTypes.className,
+  wrapperStyle: PropTypes.object,
 
   /*
    * Initial selected value.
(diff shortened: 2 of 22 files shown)


package.json
```diff
@@ -93,7 +93,8 @@
     "interact.js": "^1.2.6",
     "lodash": "^4.5.0",
     "react": "^15.0.1",
-    "react-addons-shallow-compare": "^15.0.2",
+    "react-addons-pure-render-mixin": "^15.1.0",
+    "react-addons-shallow-compare": "^15.1.0",
```
GabeMedrash: is this being used?

src/ui/slider/src/slider.jsx
```diff
@@ -238,82 +160,173 @@ export default class Slider extends React.Component {
...
+   *   _.identity
+   */
+  labelFunc: PropTypes.func,
+
+  /* include fill in the track to indicate value. */
+  fill: PropTypes.bool,
 
-Slider.defaultProps = defaultProps;
+  /* style for the fill color. */
+  fillColor: PropTypes.string,
```
GabeMedrash: i think this should be changed to `fillStyle` and accept an object; then we can take fillStyle out of state

src/ui/slider/src/track.jsx
```diff
@@ -60,12 +63,33 @@ export default class Track extends React.Component {
 
   render() {
     return (
-      <div className={style.track}>
-        <div ref={this.trackRef} className={style['track-click-target']}></div>
+      <div className={classNames(style.track, this.props.className)} style={this.props.style}>
         {this.props.children}
+        {this.state.ticks && (
+          <Ticks
+            className={this.props.ticksClassName || style['track-ticks']}
```
GabeMedrash: what about `className={classNames(style['track-ticks], this.props.ticksClassName)}`? that way styles provided via props can build off any base provided here.

src/ui/slider/src/track.jsx
```diff
@@ -60,12 +63,33 @@ export default class Track extends React.Component {
...
+  style: PropTypes.object,
+  children: PropTypes.node,
+  onClick: PropTypes.func.isRequired,
+  snapTarget: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.object,
+  ]).isRequired,
+
+  /* show ticks */
+  ticks: PropTypes.bool,
```
GabeMedrash: this gets to the conversation we had regarding these types of boolean props--whether "ticks" or "showTicks" or "renderTicks". I don't think we decided anything, but we need to flag this as something to standardize.

src/ui/slider/test/util.test.js
```diff
@@ -0,0 +1,54 @@
+import { expect } from 'chai';
+
+import {
+  getDimension,
+  getFloatPrecision,
+  getSnapTargetFunc,
+  valueWithPrecision,
+} from '../src/util';
+
```
GabeMedrash: all of these `describes` should be nested in one that will make it easy to know where they live, i.e.:
```
describe('<Slider />', () => {
 ..
});
```

src/ui/slider/src/slider.jsx
```diff
@@ -238,82 +160,173 @@ export default class Slider extends React.Component {
...
+  ticks: PropTypes.bool,
+
+  /*
+   * callback function when value is changed.
+   * Params:
+   *   value - object with keys ['min'] and 'max'
+   *   key - key of most recent value change.
+   */
+  onChange: PropTypes.func.isRequired,
+};
```
GabeMedrash: it might be helpful to have the following props as well:
- handleClassName
- handleStyles
- fillClassName

PR #36, opened 2016-06-30
Symbol select

aaronKeel:
Now you can select and focus on scatter symbols. See demo.

### DIFF
--- a/src/ui/shape/src/symbol.jsx
+++ b/src/ui/shape/src/symbol.jsx
@@ -36,10 +36,12 @@ export default class Symbol extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (
-      nextProps.type !== this.props.type ||
-      nextProps.size !== this.props.size
-    ) {
+    /**
+     * Use propsChanged here.
+     */
+    if (propsChanged(this.props, nextProps, [
+      'type', 'size'
+    ])) {
       this.setState({
         path: this.createPath(nextProps.type, nextProps.size),
       });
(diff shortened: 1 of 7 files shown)


src/ui/shape/src/symbol.jsx
```diff
@@ -36,10 +36,12 @@ export default class Symbol extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (
-      nextProps.type !== this.props.type ||
-      nextProps.size !== this.props.size
-    ) {
+    /**
```
GabeMedrash: no need for this comment

src/ui/shape/src/symbol.jsx
```diff
@@ -52,15 +54,30 @@ export default class Symbol extends React.Component {
     ]);
   }
 
-  getStyle() {
+  /**
+   * pass in data for selectedStyle(and other style) for function.
+   * return a new object using Object.assign() instead of return for each if.
+   * Reorder styling, default -> selected -> focused.
+   */
+  getStyle(data) {
```
GabeMedrash: nice job here. we need to use lodash::assign instead of Object.assign for browser compatibility though.

src/ui/shape/src/symbol.jsx
```diff
@@ -52,15 +54,30 @@ export default class Symbol extends React.Component {
     ]);
   }
 
-  getStyle() {
+  /**
+   * pass in data for selectedStyle(and other style) for function.
+   * return a new object using Object.assign() instead of return for each if.
+   * Reorder styling, default -> selected -> focused.
```
GabeMedrash: This is nitpicky, but comment should be about what the function does, not what the refactoring process captured. E.g., "Reorder styling, default -> selected -> focused" should be something more like "Styling is applied in the following order...".

src/ui/shape/src/symbol.jsx
```diff
@@ -98,7 +115,7 @@ export default class Symbol extends React.Component {
         onMouseLeave={eventHandleWrapper(onMouseLeave, data, this)}
         onMouseMove={eventHandleWrapper(onMouseMove, data, this)}
         onMouseOver={eventHandleWrapper(onMouseOver, data, this)}
-        style={this.getStyle()}
+        style={this.getStyle(data)}
```
GabeMedrash: Because this will create a new object on each call, I think we should move the creation of the inline-styles into state. The optimization will work exactly the same as the optimization you completed for caching the path. [Strictly speaking](https://github.com/facebook/react/blob/master/src/renderers/dom/client/ReactMount.js#L447) I don't think it will matter for performance, but the intelligent caching of objects as opposed to creating new ones on each render is probably a good thing.

GabeMedrash:
Merging this so that I can pick it up.

PR #38, opened 2016-07-07
Option button enhancements

brettvitaz:
This resolves an issue with `Group` and `Option`s not showing correct styling because `Button` was stomping on them.

Cleaned up `className` handling.
Added new `className` and `style` props for `selected` and `disabled`.
Made `Group`, `Option`, and `Button` extend `PureComponent` for efficient rendering.
Tried to maintain backwards compatibility as much as possible.

@davschne: Please review this and make sure it suits your needs

### DIFF
--- a/src/ui/group/src/group.jsx
+++ b/src/ui/group/src/group.jsx
@@ -5,13 +5,19 @@ import { CommonPropTypes, PureComponent } from '../../../utils';
 
 import styles from './group.css';
 
-const wrappedClickHandler = memoize((wrappedProps, clickHandler) => {
-  return () => {
-    clickHandler({ value: wrappedProps });
-  };
-});
-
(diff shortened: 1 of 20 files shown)


src/ui/group/src/group.jsx
```diff
@@ -1,76 +1,49 @@
...
-  ]).isRequired,
-
-  className: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.object,
-    PropTypes.array
-  ]),
+const wrappedClickHandler = memoize((wrappedProps, clickHandler) => {
+  return () => {
+    clickHandler({ value: wrappedProps });
```
brettvitaz: I'd like to see this change to not return an object, but I don't want to break compatibility. 
Let me know what you think.

PR #40, opened 2016-08-03
feature/tooltip

GabeMedrash:
This is a base implementation of a tooltip component that can wrap any arbitrary markup/component.

### DIFF
--- a/src/ui/tooltip/src/tooltip.jsx
+++ b/src/ui/tooltip/src/tooltip.jsx
@@ -9,8 +9,8 @@ export default class Tooltip extends PureComponent {
   /**
    * @param {Object} params
    * @param {Number} params.height - height of tooltip
-   * @param {Number} params.mouseClientX
-   * @param {Number} params.mouseClientY
+   * @param {Number} params.mouseX
+   * @param {Number} params.mouseY
    * @param {Number} params.offsetX
    * @param {Number} params.offsetY
    * @param {Number} params.paddingX
@@ -22,8 +22,8 @@ export default class Tooltip extends PureComponent {
    */
   static getPosition({
     height,
-    mouseClientX,
-    mouseClientY,
+    mouseX,
+    mouseY,
     offsetX,
     offsetY,
     paddingX,
@@ -34,19 +34,19 @@ export default class Tooltip extends PureComponent {
   }) {
     // aim to position tooltip centered about the mouse-cursor in the x-direction
     const halfWidth = width / 2;
-    const offsetXCoordinate = mouseClientX + offsetX;
+    const offsetXCoordinate = mouseX + offsetX;
 
     let x = offsetXCoordinate - halfWidth;
     // guard against placing the tooltip off the left-side of the screen
     if (x < paddingX) x = offsetXCoordinate + paddingX;
     // guard against placing the tooltip off the right-side of the screen
     if (offsetXCoordinate + halfWidth > windowInnerWidth - paddingX) {
-      x = mouseClientX - width - paddingX - offsetX;
+      x = mouseX - width - paddingX - offsetX;
     }
 
     // position tooltip above or below the mouse cursor in the y-direction
     // origin in top-left corner
-    const offsetYCoordinate = mouseClientY - offsetY;
+    const offsetYCoordinate = mouseY - offsetY;
     let y = (offsetY < 0)
             ? offsetYCoordinate // assume tooltip should be placed below mouse
             : offsetYCoordinate - height; // otherwise, assume tooltip should be placed above mouse
@@ -55,7 +55,7 @@ export default class Tooltip extends PureComponent {
     if (y < paddingY) y = offsetYCoordinate + paddingY;
     // guard against placing tooltip below screen
     if (offsetY < 0 && offsetYCoordinate + height > windowInnerHeight - paddingY) {
-      y = mouseClientY - height - paddingY - Math.abs(offsetY);
+      y = mouseY - height - paddingY - Math.abs(offsetY);
     }
 
     return { transform: `translate(${x}px, ${y}px)` };
@@ -88,8 +88,8 @@ export default class Tooltip extends PureComponent {
 
   componentWillReceiveProps(nextProps) {
     if (!propsChanged(this.props, nextProps, [
-      'mouseClientX',
-      'mouseClientY',
+      'mouseX',
+      'mouseY',
       'offsetX',
       'offsetY',
       'paddingX',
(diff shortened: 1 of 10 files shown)


davschne:
Looks good, @GabeMedrash!

src/ui/tooltip/src/tooltip.jsx
```diff
@@ -0,0 +1,229 @@
...
+   */
+  static getBounds(bounds) {
+    return {
+      x: getValue(bounds, 'x', [0, getValue(window, 'innerWidth', 0)]),
+      y: getValue(bounds, 'y', [0, getValue(window, 'innerHeight', 0)]),
+    };
+  }
+
+  /**
+   * @param {Object} params
```
brettvitaz: I haven't found good documentation for this yet, but in my opinion, destructured arguments don't need the `params` object, as you have put here. I would go so far as to say it's misleading to include. My suggestion is to just remove `params` and the reference from these params. Intellij seems to prefer it that way anyway...

src/ui/tooltip/src/tooltip.jsx
```diff
@@ -0,0 +1,229 @@
...
+   * @param {Number} params.mouseX
+   * @param {Number} params.mouseY
+   * @param {Number} params.offsetX
+   * @param {Number} params.offsetY
+   * @param {Number} params.paddingX
+   * @param {Number} params.paddingY
+   * @param {Number} params.width - width of tooltip
+   * @return {{ left: number, top: number }}
+   */
+  static getPosition({
```
brettvitaz: Visually, this is a little difficult to distinguish from normal params. I'm OK with this, but it may be better to destructure the object inside the function.

src/ui/tooltip/src/tooltip.jsx
```diff
@@ -0,0 +1,229 @@
...
+    paddingY,
+    width,
+  }) {
+    const {
+      x: [leftBound, rightBound],
+      y: [topBound, bottomBound],
+    } = bounds;
+
+    // aim to position tooltip centered about the mouse-cursor in the x-direction
+    const halfWidth = width / 2;
```
brettvitaz: I think you can simplify all of the X logic like this:
```javascript
x = clamp(offsetX + mouseX - halfWidth, leftBound + paddingX, rightBound - paddingX - width);
```

Y logic may possibly be similarly simplified.

src/ui/tooltip/src/tooltip.jsx
```diff
@@ -0,0 +1,229 @@
...
+export default class Tooltip extends PureComponent {
+  /**
+   * Return bounds passed through props or default
+   * to { x: [0, window.innerWidth], y: [0, window.innerHeight] }
+   * @param {Object} [bounds]
+   * @param {Array} [bounds.x]
+   * @param {Array} [bounds.y]
+   * @return {{ x: array, y: array}} Bounds object with keys x and y
+   */
+  static getBounds(bounds) {
```
brettvitaz: Can this be a useful utility function?

src/ui/tooltip/src/tooltip.jsx
brettvitaz: Since this is specialized, it might not need to be a static method. Possibly just pass in the wrapper and nextProps.

src/ui/tooltip/src/tooltip.jsx
```diff
@@ -0,0 +1,229 @@
...
+   * @param {Array} params.bounds.y - default to [0, window.innerHeight]
+   * @param {Number} params.height - height of tooltip
+   * @param {Number} params.mouseX
+   * @param {Number} params.mouseY
+   * @param {Number} params.offsetX
+   * @param {Number} params.offsetY
+   * @param {Number} params.paddingX
+   * @param {Number} params.paddingY
+   * @param {Number} params.width - width of tooltip
+   * @return {{ left: number, top: number }}
```
brettvitaz: JSDoc is incorrect

brettvitaz:
thumbsup

src/ui/tooltip/src/tooltip.jsx
davschne: @brettvitaz, I think this is the correct style for a JSDoc parameter object with specific required properties. See:
http://usejsdoc.org/tags-param.html#parameters-with-properties

brettvitaz: Thanks for the link. The specific documentation I am looking for would cover destructuring argument objects in the function signature.

davschne: Ah, I guess I'm misunderstanding. Is your comment related to JSDoc style or to JS coding style? If the former, I can't see how it could matter that the parameter object is destructured in the signature; this would seem to be an internal stylistic detail that should be totally invisible to the caller.

src/ui/tooltip/src/tooltip.jsx
davschne: I like the elegance of this logic, but it's actually rather different behavior from what Gabe wrote (and it causes one of the unit tests to fail). In Gabe's version, the offset is always applied, but if the tooltip would be placed outside the bounds, the offset direction is flipped.

src/ui/tooltip/src/tooltip.jsx
davschne: @brettvitaz, can you clarify, please? Incorrect in what way?

brettvitaz: @davschne Check the return statement.

src/ui/tooltip/src/tooltip.jsx
brettvitaz: The implementation details are invisible to the caller, but the JSDoc is not correct. There is no real parameter here named `params`. There can also be multiple destructured parameters. This code is not a big problem, but it is something to consider.

src/ui/tooltip/src/tooltip.jsx
brettvitaz: It is a little different in behavior, so I would expect it to break a unit test or two.

 In my version, the offset is also always applied, though it is not flipped. The tool tip hugs the padding when it would otherwise be outside of bounds. Have you tried it with the demo?

src/ui/tooltip/src/tooltip.jsx
davschne: Okay, I see what you're saying now. Personally, I think the name of a parameter is not particularly important in JavaScript, since the language has no mechanism for dealing with keyword arguments - they're totally positional. I think the name is only important insofar as it describes the meaning of the parameter. And in order to use this style of parameter documentation, we need a name of some sort.

src/ui/tooltip/src/tooltip.jsx
brettvitaz: We have determined in our code review that this change represents the desired behaviour.

src/ui/tooltip/src/tooltip.jsx
brettvitaz: Its name is not particularly important, but it's not correct to say it has a name.

PR #42, opened 2016-08-08
Enhancement/slider backing list

brettvitaz:
To mitigate floating point math type errors, `<Slider />` will always either generate a backing array created from `range` prop object, or accept an array as the `range` prop. This change replaces  `minValue` and `maxValue` with the `range` object.
`min` and `max` have been replaced with `low` and `high`.

`range` will take the following form:
```javascript
range: {
    low: // range extent low
    high: // range extent high
    steps: // number of steps between extents (optional, if low/high are `int`)
    precision: // precision to round values (optional, if low/high are `int` and steps is omitted)
}
```
or
```javascript
range: [1990, 1991, 1992, ...]
```

### DIFF
--- a/src/ui/slider/src/slider.jsx
+++ b/src/ui/slider/src/slider.jsx
@@ -340,8 +340,8 @@ Slider.propTypes = {
   range: PropTypes.oneOfType([
     PropTypes.array,
     PropTypes.shape({
-      low: PropTypes.number,
-      high: PropTypes.number,
+      low: PropTypes.number.isRequired,
+      high: PropTypes.number.isRequired,
       steps: PropTypes.number,
       precision: PropTypes.number,
     }),
@@ -369,8 +369,8 @@ Slider.propTypes = {
     PropTypes.string,
     PropTypes.array,
     PropTypes.shape({
-      low: PropTypes.number,
-      high: PropTypes.number,
+      low: PropTypes.any,
+      high: PropTypes.any,
     }),
   ]).isRequired,
 
@@ -409,6 +409,7 @@ Slider.propUpdates = {
     return { indexes: getIndexesForValues(getLowHighValues(nextProp), state.range) };
   }),
   width: updateFunc(() => {
+    // Track needs to be rendered with new width before Handles and Fill can be rendered
     return { render: false };
   }),
 };
--- a/src/ui/slider/src/slider.jsx
+++ b/src/ui/slider/src/slider.jsx
@@ -180,9 +180,12 @@ export default class Slider extends PureComponent {
 
   updateValueFromEvent(event, index, key) {
     if (index !== this.state.indexes[key] && this.state.range[index]) {
-      const values = getValuesForIndexes({ ...this.state.indexes, [key]: index }, this.state.range);
+      const indexes = { ...this.state.indexes, [key]: index };
+      const values = getValuesForIndexes(indexes, this.state.range);
 
-      if (values.high === undefined || values.low <= values.high) {
+      if (indexes.low === undefined ||
+            indexes.high === undefined ||
+            indexes.low <= indexes.high) {
         this.props.onChange(event, values, this);
       }
     }
--- a/src/ui/slider/src/slider.jsx
+++ b/src/ui/slider/src/slider.jsx
@@ -26,10 +26,15 @@ import style from './style.css';
  */
 function getIndexesForValues(values, rangeList) {
   return reduce(values, (acc, value, key) => {
+    /* eslint-disable no-nested-ternary */
+    // if value is not in the range list, assign either first or last element, depending on key
+    const index = rangeList.indexOf(value);
     return {
       ...acc,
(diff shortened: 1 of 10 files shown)


src/ui/slider/src/slider.jsx
```diff
@@ -119,45 +160,47 @@ export default class Slider extends PureComponent {
...
-        value >= this.props.minValue &&
-        value <= this.props.maxValue) {
-      const values = { ...this.state.values, [key]: value };
+  updateValueFromEvent(index, key) {
+    if (index !== this.state.indexes[key] && this.state.range[index]) {
+      const values = getValuesForIndexes({ ...this.state.indexes, [key]: index }, this.state.range);
 
-      if (values.max === undefined || values.min <= values.max) {
+      if (values.high === undefined || values.low <= values.high) {
         this.props.onChange(valueByHandleCount({ ...values }, this.handleCount), key);
```
GabeMedrash: To be consistent with other components, I think `props.onChange` should have the following signature:
`function(event, value, Handle) {...}`. Here, `value` might be an object with { key: String, value: Any }.

src/ui/slider/src/slider.jsx
```diff
@@ -322,20 +375,38 @@ Slider.propTypes = {
...
 Slider.defaultProps = {
   width: 200,
-  step: 1,
   labelFunc: identity,
 };
 
 Slider.propUpdates = {
-  value: updateFunc((nextProp) => {
-    return { values: getMinMaxValues(nextProp) };
+  range: updateFunc((nextProp, propName, nextProps, state) => {
```
GabeMedrash: I think there's a little bit of indirection going on in `propUpdates`. I would have expected the keys of `Slider.propUpdates` to map directly to state (even though the function allows for that to not be the case), so it took me some time to notice that `indexes` is being set under `value`, `render` under `width`, and `scale` under `range`.

src/ui/slider/src/slider.jsx
```diff
@@ -322,20 +375,38 @@ Slider.propTypes = {
...
+    let nextRange = nextProp;
+
+    if (!Array.isArray(nextProp)) {
```
GabeMedrash: Would be nice to either have a more comprehensive validator (to make certain props.range is in the shape you expect), or make the props.range shape properties required.

src/ui/slider/src/slider.jsx
```diff
@@ -322,20 +375,38 @@ Slider.propTypes = {
...
+    return {
+      range: nextRange,
+      scale: state.scale.domain([0, nextRange.length - 1]),
+    };
+  }),
+  value: updateFunc((nextProp, propName, nextProps, state) => {
+    return { indexes: getIndexesForValues(getLowHighValues(nextProp), state.range) };
+  }),
+  width: updateFunc(() => {
+    return { render: false };
```
GabeMedrash: Document why setting render = false when width changes

src/ui/slider/src/slider.jsx
```diff
@@ -257,33 +305,38 @@ export default class Slider extends PureComponent {
 }
 
 Slider.propTypes = {
```
GabeMedrash: Let's alphabetize these

src/ui/slider/src/slider.jsx
```diff
@@ -1,29 +1,70 @@
...
  */
-function getMinMaxValues(value) {
-  if (typeof value === 'number') {
-    return { min: value };
-  } else if (Array.isArray(value)) {
-    return zipObject(['min', 'max'], value);
+function getLowHighValues(value) {
+  if (Array.isArray(value)) {
+    return zipObject(['low', 'high'], value);
+  } else if (typeof value === 'object') {
```
GabeMedrash: Validate that object is of correct shape, otherwise throw descriptive error

src/ui/slider/src/slider.jsx
```diff
@@ -322,20 +375,38 @@ Slider.propTypes = {
...
-  value: updateFunc((nextProp) => {
-    return { values: getMinMaxValues(nextProp) };
+  range: updateFunc((nextProp, propName, nextProps, state) => {
+    let nextRange = nextProp;
+
+    if (!Array.isArray(nextProp)) {
+      const delta = nextProp.high - nextProp.low;
+      const steps = nextProp.steps || delta + 1;
+      nextRange = range(steps).map(
+        (d) => round(d * delta / (steps - 1) + nextProp.low, nextProp.precision)
```
GabeMedrash: Check for isNil(nextProp.precision)

GabeMedrash:
Looks good--some comments inline. Some basic initial tests that I think will be important:
1. Test the range input to the slider (takes an array, creates the expected slider; takes an object of the defined shape, creates the expected slider)/
2. Test the value input to the slider.
3. Basic validation test that the onChange prop is called with `event, value, this`.

brettvitaz:
I'm going to hold off on splitting up the `styles`.

PR #44, opened 2016-08-11
Enhancement/choropleth updates for disputed territories

GabeMedrash:
- Add clip extent to `<Choropleth />`'s path generator as an easy-win (huge) performance enhancement.
- Change `<Choropleth />` prop name 'geoJSONKeyField' to 'geometryKeyField' so that it makes more sense. Also, change it to more accurately reflect that it accepts but a string and a function.
- Change `<Path />` prop 'locationId' to accept both a string and a number, because we have a use-case for both.


GabeMedrash:
This work will just be included with the upcoming compositions dump.

PR #49, opened 2016-08-24
Enhancement/spinner style

brettvitaz:
This is a style update, kind of fresh. 
Uses `svg` instead of `span`. Adds style prop. Size of spinner is determined by `font-size` style now, not `size` prop and `vw`. This makes it more consistently sized in different views.

### DIFF
--- a/src/ui/spinner/src/spinner.jsx
+++ b/src/ui/spinner/src/spinner.jsx
@@ -7,22 +7,30 @@ import styles from './spinner.css';
 
 class Spinner extends PureComponent {
   render() {
-    const { props } = this;
+    const {
+      className,
+      inline,
+      style,
+      symbolFill,
+      symbolStyle,
+      symbolType,
+    } = this.props;
+
     return (
       <div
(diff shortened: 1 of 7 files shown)


src/ui/spinner/src/spinner.jsx
```diff
@@ -1,37 +1,42 @@
...
+    const { props } = this;
+    return (
+      <div
+        className={classNames(styles.spinner, {
+          [styles['inline-spinner']]: props.inline
+        }, props.className)}
+        style={props.style}
+      >
+        <svg className={styles.dot} viewBox="-8 -8 16 16" width="1em" height="1em">
+          <Symbol symbolType="circle" fill="black" />
```
GabeMedrash: expose way for symbol's to not be black

brettvitaz:
@GabeMedrash I cleaned up the props destructuring and added the ability to modify the symbol style.

GabeMedrash:
👍

PR #52, opened 2016-08-31
Feature/multi scatter consistency

brettvitaz:
This PR makes the MultiScatter more closely match the recent MultiLine enhancements.

Highlights:
* Extend `PureComponent` class
* Add scatter value Iteratee function


src/ui/shape/src/multi-scatter.jsx
```diff
@@ -164,10 +183,7 @@ MultiScatter.defaultProps = {
   colorScale() { return 'steelblue'; },
   dataField: 'values',
   keyField: 'key',
-  onClick: noop,
-  onMouseLeave: noop,
-  onMouseMove: noop,
-  onMouseOver: noop,
```
aaronKeel: Do mouse events still have a default? CommonDefaultProps.noop?

brettvitaz: `Scatter` implements the same defaults, so IMO it is unnecessary to re-declare them here.

aaronKeel:
"Thumbs up."

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -77,9 +88,9 @@ MultiScatter.propTypes = {
      y: property on data to position in y-direction
    */
   dataAccessors: PropTypes.shape({
-    fill: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
-    x: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
-    y: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+    fill: CommonPropTypes.dataAccessor,
+    x: CommonPropTypes.dataAccessor.isRequired,
```
GabeMedrash: i don't think either x or y should be required--what if you want multiple scatter plots of a single dimension?

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -125,6 +136,14 @@ MultiScatter.propTypes = {
    */
   scatterClassName: CommonPropTypes.className,
 
+  /*
+   function to apply to the datum to transform area values. default: _.identity
```
GabeMedrash: should be "...to transform scatter values...", right?

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -1,63 +1,74 @@
...
+      'focusedStyle',
+      'selectedClassName',
+      'selectedStyle',
+      'scales',
+      'size',
+      'style',
+      'symbolClassName',
+    ]);
+
+    const mouseEvents = pick(this.props, [
```
GabeMedrash: why break these out of `childProps`?

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -1,63 +1,74 @@
...
+    ]);
+
+    return (
+      <g className={className && classNames(className)}>
+        {
+          map(data, (datum) => {
+            const key = propResolver(datum, keyField);
+            const values = propResolver(datum, dataField);
+            const symbol = propResolver(datum, symbolField);
+            const fill = colorScale(key);
```
GabeMedrash: in the spirit of having split out `dataAccessors.fill` and `dataAccessors.key`, what about accepting a new prop `fillField` that can be used like:
```
const fillValue = propResolver(datum, fillField || keyField);
const fill = colorScale(fillValue);
```

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -1,63 +1,74 @@
...
+            const symbolType = symbolScale(symbol);
+
+            const scatterValues = scatterValuesIteratee(values, key);
```
GabeMedrash: would be nice to have a test that does a simple check--if scatterValuesIteratee returns falsey, does not produce an associated scatter, if truthy, does produce an asssociated scatter

GabeMedrash:
looking good

PR #53, opened 2016-08-31
add clipPath prop to axisChart

aaronKeel:
These changes give axis-chart a new prop `clipPath [boolean]` that can implement a `<defs>` containing a rectangular clipPath with width and height of the chart display area. Thus if a line, area, or symbol extends outside the clipPath, it is not painted on the svg (use case is for scaling axes, changing domain or range).

Shape components all receive a `clipPathId [string]` prop that is used by the primitive component to access the clipPath.


GabeMedrash:
Looks good, @aaronKeel. Ready to merge.

PR #57, opened 2016-09-07
enhancement/number formatter

GabeMedrash:
default to formatting numbers to the thousandths place, but now accept a precision parameter

### DIFF
--- a/src/utils/numbers.js
+++ b/src/utils/numbers.js
@@ -4,11 +4,12 @@ import { isFinite } from 'lodash';
  * number formatter to be used as, for example, tickFormat fn and label formatter
  * if value is a non-finite number, returns empty string
  * @param {number} value
+ * @param {number} [precision] defaults to three decimal places
  * @return {string}
  */
-export function numberFormat(value) {
+export function numberFormat(value, precision = 3) {
   if (!isFinite(value)) return '';
-  if (value >= 1000) return `${(value / 1000).toFixed(3)}k`;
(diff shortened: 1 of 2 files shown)


src/utils/numbers.js
```diff
@@ -4,11 +4,12 @@ import { isFinite } from 'lodash';
  * number formatter to be used as, for example, tickFormat fn and label formatter
  * if value is a non-finite number, returns empty string
  * @param {number} value
+ * @param {number} [precision] defaults to three decimal places
  * @return {string}
  */
-export function numberFormat(value) {
+export function numberFormat(value, precision = 3) {
   if (!isFinite(value)) return '';
```
GabeMedrash: @brettvitaz I'm curious what you think about this guard statement. If the value passed in is not a finite number, should we just return whatever that value happens to be, or keep as is and return an empty string?

brettvitaz: Have you looked at d3's number formatter library? This might be useful (if you haven't seen it):
https://github.com/d3/d3-format#precisionPrefix

GabeMedrash: I considered using `d3-format`. I'm not really sold on the formatting specification "mini-language" though--it looks powerful, but I think the cost paid in having little idea what it does by just reading it is too high for our simple use cases.

brettvitaz: My suggestion is to add an argument that allows the user to return non-finite number strings, and set to True by default

PR #62, opened 2016-09-23
fix/expandables style

GabeMedrash:
"expand/contract" icon modified for:
- consistency with ihme-ui button
- bigger

### DIFF
--- a/src/ui/expansion-container/src/Expandable.jsx
+++ b/src/ui/expansion-container/src/Expandable.jsx
@@ -238,7 +238,7 @@ export default class Expandable extends PureComponent {
         style={iconStyle}
         onClick={this.wrappedEvent((expanded && this.restore) || (restored && this.expand))}
         viewBox="-16 -16 32 32"
-        width="1em" height="1em"
+        width="20px" height="20px"
       >
         <circle
           r="15"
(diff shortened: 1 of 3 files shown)


src/ui/expansion-container/src/Expandable.jsx
```diff
@@ -229,28 +225,25 @@ export default class Expandable extends PureComponent {
...
 
     return (
       <svg
-        className={iconClassName}
+        className={classNames(styles['icon-container'], iconClassName)}
         style={iconStyle}
         onClick={this.wrappedEvent((expanded && this.restore) || (restored && this.expand))}
         viewBox="-16 -16 32 32"
-        width="1em" height="1em"
+        width="20px" height="20px"
```
brettvitaz: IMO, this should remain `1em` and `props.iconStyle` set to `{ fontSize: '20px' }`.

The transitions are still based on `em`s, so if we change this, we need to change the transitions as well.

src/ui/expansion-container/src/expansion-container.css
```diff
@@ -16,6 +16,7 @@
 
 .expandable-container {
   position: relative;
+  padding-right: 20px;
```
brettvitaz: This should be on `expandable-content`, and probably be `1em`. It can be overridden by assigning `prop.expandableStyle`

PR #65, opened 2016-11-03
accept list of topojson objects to turn into map layers

GabeMedrash:
- replaces `subnational` prop with list of topojson objects to render
- a breaking change, but will default to `topojsonObjects=['national']`, which will at least preserve initial loading behavior in Mort/CodViz


src/ui/compositions/map/src/map.jsx
```diff
@@ -623,26 +600,24 @@ Map.propUpdates = {
...
-        if (layer.name === 'selected-non-disputed-national-borders'
-            || layer.name === 'selected-non-disputed-subnational-borders') {
+        if (endsWith(layer.name, 'selected-non-disputed-borders')) {
           meshFilter = context.selectedBordersMeshFilter(nextProps.selectedLocations);
         }
         return assign({}, layer, { filterFn: meshFilter });
       }),
     });
   },
   subnational: (state, _, prevProps, nextProps, context) => {
```
aaronKeel: Should this be called topojsonObjects?

src/ui/compositions/map/src/map.jsx
```diff
@@ -533,24 +509,24 @@ Map.propTypes = {
...
 
   titleClassName: CommonPropTypes.className,
 
   titleStyle: PropTypes.object,
 
   /*
+    array of keys on topology.objects (e.g., 'national', 'ADM1', 'health_districts');
+    if a key on topology.objects is omitted, it will not be rendered
+  */
+  topojsonObjects: PropTypes.arrayOf(PropTypes.string),
```
aaronKeel: Submitting a list of layer names looks like a good idea for now. We may want to check performance for topojsonObjects that do not have disputed borders. Also should there be a guard against including topojsonObject names that do not exist in the topojson.objects? The issue with either case is having unnecessary layers.

src/ui/compositions/map/src/map.jsx
```diff
@@ -284,6 +220,46 @@ export default class Map extends React.Component {
     };
   }
 
+  createLayers(name) {
```
aaronKeel: Since all layers are `visible: true` can `getGeometryIds` be simplified on line 156?

src/ui/compositions/map/src/map.jsx
```diff
@@ -623,26 +600,24 @@ Map.propUpdates = {
     if (isEqual(nextProps.selectedLocations, prevProps.selectedLocations)) return state;
     return assign({}, state, {
       layers: state.layers.map(layer => {
+        if (layer.type !== 'mesh') return layer;
+
         let meshFilter = layer.filterFn;
-        if (layer.name === 'selected-non-disputed-national-borders'
-            || layer.name === 'selected-non-disputed-subnational-borders') {
+        if (endsWith(layer.name, 'selected-non-disputed-borders')) {
```
brettvitaz: Check to see if this is necessary.

PR #66, opened 2016-12-05
enhancement/select

GabeMedrash:
move expressions from `render` that produce new objects into state


src/ui/select/src/multi-select.jsx
```diff
@@ -2,15 +2,15 @@ import React, { PropTypes } from 'react';
 import Select, { propTypes as baseProps } from 'ihme-react-select';
 import { assign } from 'lodash';
 
-import { stateFromPropUpdates, propsChanged } from '../../../utils';
+import { stateFromPropUpdates, propsChanged, PureComponent } from '../../../utils';
```
aaronKeel: Does it make a difference if we use PureComponent from  utils, or the React.PureComponent as stated in
[https://facebook.github.io/react/docs/pure-render-mixin.html](https://facebook.github.io/react/docs/pure-render-mixin.html)

and [https://facebook.github.io/react/docs/react-api.html#react.purecomponent](https://facebook.github.io/react/docs/react-api.html#react.purecomponent)
> React.PureComponent is exactly like React.Component but implements shouldComponentUpdate() with a shallow prop and state comparison.

maybe this has already been addressed.

src/ui/select/src/single-select.jsx
```diff
@@ -2,13 +2,13 @@ import React, { PropTypes } from 'react';
 import Select, { propTypes as baseProps } from 'ihme-react-select';
 import { assign } from 'lodash';
 
-import { stateFromPropUpdates, propsChanged } from '../../../utils';
+import { stateFromPropUpdates, propsChanged, PureComponent } from '../../../utils';
```
aaronKeel: Same deal as multi-select.

Does it make a difference if we use PureComponent from utils, or the React.PureComponent as stated in
https://facebook.github.io/react/docs/pure-render-mixin.html

and https://facebook.github.io/react/docs/react-api.html#react.purecomponent

React.PureComponent is exactly like React.Component but implements shouldComponentUpdate() with a shallow prop and state comparison.
maybe this has already been addressed.

src/ui/select/src/multi-select.jsx
```diff
@@ -2,78 +2,112 @@ import React, { PropTypes } from 'react';
...
+  /* width added to widest label (in px) */
+  widthPad: PropTypes.number,
+};
+
 MultiSelect.propTypes = assign({}, baseProps, multiSelectPropTypes);
-MultiSelect.defaultProps = defaultProps;
+
+MultiSelect.defaultProps = {
+  placeholder: 'Add/remove',
+  resetValue: [],
```
aaronKeel: I like the idea of putting more defaults here. Good practice.

src/ui/select/src/multi-select.jsx
```diff
@@ -2,78 +2,112 @@ import React, { PropTypes } from 'react';
...
         valueRenderer={multiValueRenderer}
-        menuStyle={assign({}, { overflow: 'hidden', width: `${menuWidth}px` }, menuStyle)}
-        menuContainerStyle={assign({}, { width: `${menuWidth}px` }, menuContainerStyle)}
-        resetValue={resetValue || []}
+        menuStyle={menuStyle}
+        menuContainerStyle={menuContainerStyle}
+        resetValue={resetValue}
         autosize={false}
         {...this.props}
       />
```
aaronKeel: I definitely like paying more attention to where and when new objects are created. Moving `assign` out of react components definitions makes it easier to read.

src/ui/select/src/multi-select.jsx
GabeMedrash: Well, it shouldn't make a difference, and I agree that we should start using `React.PureComponent`, but that was released in v15.3.0, so I think we should (in a separate work effort) update React, make sure nothing is broken, then excise our home-rolled "PureComonent" in favor of what the library provides.

PR #79, opened 2017-01-05
Menu up

aaronKeel:
I added a property `menuUpward` to SingleSelect and MultiSelect. This also includes a `FLIP_MENU_UPWARDS_INLINE_STYLE` constant and a `getMenuContainerStyle` function in the utils. You can verify its functionality in the DEMO. Let me know if everything is pizza, or if poop.

### DIFF
--- a/src/ui/select/src/multi-select.jsx
+++ b/src/ui/select/src/multi-select.jsx
@@ -4,7 +4,7 @@ import Select, { propTypes as baseProps } from 'ihme-react-select';
 import { assign } from 'lodash';
 
 import { stateFromPropUpdates, propsChanged, PureComponent } from '../../../utils';
-import { getMenuContainerStyle, getWidestLabel } from './utils';
+import { FLIP_MENU_UPWARDS_INLINE_STYLE, getWidestLabel } from './utils';
 
 import style from './select.css';
 import { menuWrapper } from './menu';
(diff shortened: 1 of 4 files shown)


src/ui/select/src/utils/index.js
```diff
@@ -25,3 +25,17 @@ export const getWidestLabel = (
     return Math.max(maxWidth, fullWidth);
   }, 0);
 };
+
+export const FLIP_MENU_UPWARDS_INLINE_STYLE = {
```
brettvitaz: I suggest that you add this to the `select.css` file and use `classNames` to conditionally apply the classname.

src/ui/select/src/single-select.jsx
```diff
@@ -84,7 +87,7 @@ SingleSelect.propUpdates = {
     return assign({}, state, {
       menuContainerStyle: assign({}, {
         width: `${menuWidth}px`,
-      }, nextProps.menuContainerStyle),
+      }, nextProps.menuContainerStyle, getMenuContainerStyle(nextProps.menuUpward)),
```
brettvitaz: Using a css selector, you can implement this on line 37 and forego the utility function.

src/ui/select/src/multi-select.jsx
```diff
@@ -94,7 +97,7 @@ MultiSelect.propUpdates = {
     return assign({}, state, {
       menuContainerStyle: assign({}, {
```
GabeMedrash: we should add `menuUpward` to the list of props we're checking for to see if this needs to run, i.e., the 
`if(!propsChanged(...))` guard.

src/ui/select/src/single-select.jsx
```diff
@@ -84,7 +87,7 @@ SingleSelect.propUpdates = {
     return assign({}, state, {
```
GabeMedrash: same comment here as in MultiSelect

src/ui/select/src/utils/index.js
```diff
@@ -25,3 +25,17 @@ export const getWidestLabel = (
...
+  borderRadius: '4px 4px 0 0',
+  bottom: '100%',
+  marginTop: '0px',
+  marginBottom: '-1px',
+  top: 'auto',
+};
+
+export function getMenuContainerStyle(menuUpward) {
```
GabeMedrash: I'm not sure how necessary this function is. Instead of calling this function, could be simplified to:`nextProps.menuUpdwards && FLIP_MENU_UPWARDS_INLINE_STYLE`. `lodash.assign` will handle this properly.

brettvitaz:
:pizza:

PR #84, opened 2017-01-11
Issue 69- changed Axis label Proptype from any to string

kbeame:
Changed the Axis label propType from "any" to "string".


kbeame:
@GabeMedrash Please review

Issue #89, opened 2017-01-19
SingleSelect and MultiSelect: merge components and differentiate by props


GabeMedrash:
...no reason why we can't expose a single `<Select />` component that takes a `multi` prop.

GabeMedrash:
Closed by #90

PR #90, opened 2017-02-01
Merge single multi select

aaronKeel:
#89 A new component `Select` that merges a lot of common code from `MultiSelect` and `SingleSelect`. A multi-select component is created via props by the boolean prop `multiselect`. Any suggestions on renaming the prop are welcome. `Select` defaults to a single-select. The demo has been updated to include the new `Select` component.

### DIFF
--- a/src/ui/select/index.js
+++ b/src/ui/select/index.js
@@ -1,2 +1,3 @@
 export { default as MultiSelect } from './src/multi-select';
 export { default as SingleSelect } from './src/single-select';
+export { default as Select } from './src/select';
--- a/src/ui/select/index.js
+++ b/src/ui/select/index.js
@@ -1,3 +1,9 @@
-export { default as MultiSelect } from './src/multi-select';
-export { default as SingleSelect } from './src/single-select';
-export { default as Select } from './src/select';
+import MultiSelect from './src/multi-select';
+import SingleSelect from './src/single-select';
+import Select from './src/select';
+
+export {
+  MultiSelect,
+  Select,
+  SingleSelect,
+};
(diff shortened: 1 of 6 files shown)


src/ui/select/src/select.jsx
```diff
@@ -0,0 +1,177 @@
...
+    let placeholder;
+
+    if (!this.props.placeholder) {
+      placeholder = { placeholder: 'Add/Remove' };
+    }
+
+    return (
+      <BaseSelect
+        {...this.props}
+        {...placeholder}
```
GabeMedrash: Same comment as above re:placeholder.

src/ui/select/src/select.jsx
```diff
@@ -0,0 +1,177 @@
...
+    return this.renderSingleSelect();
+  }
+}
+
+const selectPropTypes = {
+  /* drop down will flip up */
+  menuUpward: PropTypes.bool,
+
+  /* allow multiple selections */
+  multiSelect: PropTypes.bool,
```
GabeMedrash: Why not just use `props.multi`?

src/ui/select/index.js
```diff
@@ -1,2 +1,3 @@
 export { default as MultiSelect } from './src/multi-select';
 export { default as SingleSelect } from './src/single-select';
+export { default as Select } from './src/select';
```
GabeMedrash: I'd like to expose this new, merged component as the default export (`export { default as default} from './src/select';`) and add deprecation notices in `MultiSelect` and `SingleSelect` (for example, by adding a `console.warn` in their respective constructors).

PR #98, opened 2017-02-07
housekeeping

GabeMedrash:
There are two substantive changes in this PR:

**1. Export the library to support CommonJS, ES6, and UMD targets.**
CommonJS build is transpiled (using babel-preset-2015 _inclusive_ of modules transform) under `lib`; ES6 build is transpiled (using babel-preset-2015 _exclusive_ of modules transform) under `es`; UMD build is bundled with webpack 2 and put in a `dist` directory. Relevant fields in `package.json` have been updated to reflect this:  `browser` points to "dist/ihme-ui.min.js", `main` points to "lib/index.js", and both `module` and `jsnext:main` point to "es/index.js". 

There is a **major breaking change here**: the library is no longer transpiled directly into the project root, so import statements like `import AxisChart from 'ihme-ui/ui/axis-chart'` are no longer possible. Instead, every export of this library can now be imported in one of two ways: either 
`import { AxisChart } from 'ihme-ui'`, or
```
import * as ihmeUI from 'ihme-ui'; 
const AxisChart = ihmeUI.AxisChart;
```
This is desirable because it ensures consistency with how this library is imported with both webpack 1.x (which requires the use of the CommonJS build) and webpack 2.x (which can take advantage of the ES6 build). It also has the benefit of shielding library users from having to know the internal directory structure of the library in order to import a component.

**2. All demos now rely on the same webpack configuration.**
Before, all demos required their own webpack configuration file, which was often copy-pasted from one demo to another. I've refactored the demo script to use a single configuration file.


src/index.js
```diff
@@ -1,10 +1,4 @@
...
-import * as utils from './utils';
-
-module.exports = {
-  ui,
-  testUtils,
-  utils
-};
+export * from './ui';
+export * from './test-utils';
+export * from './utils';
```
brettvitaz: could this cause namespace collisions?

src/ui/legend/demo/app.jsx
```diff
@@ -76,51 +76,51 @@ class App extends React.Component {
...
-    symbolColor: 'red',
-    symbolType: 'line'
-  },
-  {
-    label: 'UN Pop',
-    symbolColor: 'grey',
-    symbolType: 'cross'
-  },
-];
+          {/* <pre><code>
```
brettvitaz: did you mean to indent this block?

src/index.js
GabeMedrash: It definitely could. Moving forward--for as long as this current setup is used--we'll have to be careful with how we name the exports of this library.

As we've discussed, we might consider breaking apart the library into smaller repos, and structuring our exports such that something like the following is possible:
```js
import { Map } from 'ihme-ui/React'
import { Layer } from 'ihme-ui/Leaflet'
```

src/ui/legend/demo/app.jsx
GabeMedrash: Nope! Good catch.

GabeMedrash:
@aaronKeel This is now merged in case you need to `git pull --rebase origin master` on your linting branch

PR #99, opened 2017-02-25
Numbers fix

aaronKeel:
Made some adjustments to make float object handle the case of 64.79 - 64.75. Added an extra buffer to `floatToIntMultiplier` and use a floor function to handle the case of 64.79 * 1000. Added tests.


GabeMedrash:
Looks good!

PR #103, opened 2017-03-08
rename Symbol to Shape

GabeMedrash:
Symbol shadows the built-in JS primitive, `Symbol`. Renaming to Shape.


src/ui/shape/test/scatter.test.js
```diff
@@ -98,7 +98,7 @@ describe('<Scatter />', () => {
       });
     };
 
-    shallow(component).find(Symbol).forEach(assertion);
+    shallow(component).find(Shape).forEach(assertion);
   });
 
   it('selects a symbol', () => {
```
aaronKeel: references to symbol in tests may cause confusion.

GabeMedrash:
@aaronKeel Thanks for the careful review! Agreed that we need to be consistent with propTypes and fix all references in comments.

PR #104, opened 2017-03-08
Allow disabling a SelectOption with option.disabled: true.

davschne:
- prevents calling props.selectValue on click
- prevents calling props.focusOption on mouseover
- prevents applying 'focused', 'selected' CSS classes
- tests included in commit

### DIFF
--- a/src/ui/select/src/select-option.jsx
+++ b/src/ui/select/src/select-option.jsx
@@ -105,8 +105,8 @@ export default class SelectOption extends React.Component {
       <div
         className={classNames(
           styles.option, {
-            [styles.focused]: !isDisabled && isFocused,
-            [styles.selected]: !isDisabled && isSelected,
+            [styles.focused]: isFocused,
+            [styles.selected]: isSelected,
           }
         )}
         onClick={!isDisabled && this.onClick}
--- a/src/ui/select/src/select-option.jsx
+++ b/src/ui/select/src/select-option.jsx
@@ -107,6 +107,7 @@ export default class SelectOption extends React.Component {
           styles.option, {
             [styles.focused]: isFocused,
             [styles.selected]: isSelected,
+            [styles.disabled]: isDisabled,
           }
         )}
         onClick={!isDisabled && this.onClick}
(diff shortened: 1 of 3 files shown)


src/ui/select/src/select-option.jsx
```diff
@@ -99,17 +99,18 @@ export default class SelectOption extends React.Component {
 
     const isFocused = option === focusedOption;
     const isSelected = valueArray ? valueArray.includes(option) : false;
+    const isDisabled = Boolean(option.disabled);
```
GabeMedrash: Why use `Boolean(...)` here?

src/ui/select/src/test/select-option.test.js
```diff
@@ -55,4 +56,33 @@ describe('<SelectOption />', () => {
       expect(wrapper.find('Input')).to.not.be.checked();
     });
   });
+
```
GabeMedrash: Great tests, thanks!

src/ui/select/src/select-option.jsx
davschne: I prefer to have a boolean type for the truthiness tests rather than a potential `undefined`. I guess it's not totally critical, but I think it's nice for a variable whose name implies it's a boolean ("is..") to actually be a boolean.

src/ui/select/src/select-option.jsx
```diff
@@ -99,17 +99,18 @@ export default class SelectOption extends React.Component {
...
 
     return (
       <div
         className={classNames(
           styles.option, {
-            [styles.focused]: isFocused,
-            [styles.selected]: isSelected,
+            [styles.focused]: !isDisabled && isFocused,
+            [styles.selected]: !isDisabled && isSelected,
```
davschne: Not a bad idea for a general use case. In my use case, though, I don't want it grayed out, so I'd need to override that style. I could do that with the optionStyle prop, right?

davschne: Just tried it; seems to work fine.

PR #111, opened 2017-04-14
Expandable hover

aaronKeel:
Mouse events are added to the expandable component so that when the mouse interacts with the button, we can provide callbacks.

### DIFF
--- a/src/ui/expansion-container/demo/app.jsx
+++ b/src/ui/expansion-container/demo/app.jsx
@@ -67,9 +67,6 @@ class App extends React.Component {
       clicks: 0,
     };
     this.onClick = this.onClick.bind(this);
-    this.onMouseLeave = this.onMouseLeave.bind(this);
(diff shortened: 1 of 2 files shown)


aaronKeel:
I also alphabetized prop lists.

src/ui/expansion-container/demo/app.jsx
```diff
@@ -61,12 +67,27 @@ class App extends React.Component {
       clicks: 0,
     };
     this.onClick = this.onClick.bind(this);
+    this.onMouseLeave = this.onMouseLeave.bind(this);
+    this.onMouseMove = this.onMouseMove.bind(this);
+    this.onMouseOver = this.onMouseOver.bind(this);
```
GabeMedrash: Minor point given that this is just a demo, but these binds are unnecessary because the methods you included don't use `this`.

PR #112, opened 2017-04-20
Feature/choropleth props management

### LINKED ISSUE #78: Choropleth: remove prop layers.visible
Layer visibility activation has changed and layers.visible is out of date.

GabeMedrash:
Props management for `<Choropleth />`:
- removes layer.visible and associated logic
- renames layer.filterFn to layer.filter to remove unnecessary verbosity 

Closes #78.

### DIFF
--- a/src/ui/choropleth/src/choropleth.jsx
+++ b/src/ui/choropleth/src/choropleth.jsx
@@ -454,7 +454,7 @@ Choropleth.propTypes = {
    * layers of topojson to include
    * layer description: {Object}
    *  - `className`: className applied to layer
-   *  - `filterFn`: optional function to filter mesh grid, passed adjacent geometries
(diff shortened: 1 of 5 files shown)


src/ui/choropleth/src/choropleth.jsx
```diff
@@ -473,7 +473,7 @@ Choropleth.propTypes = {
      * optional function to filter mesh grid, passed adjacent geometries
      * refer to https://github.com/mbostock/topojson/wiki/API-Reference#mesh
      */
-    filterFn: PropTypes.func,
+    filter: PropTypes.func,
```
aaronKeel: Would `meshFilter` be a better name since it is an optional function passed to the mesh method? `filter` by itself seems a bit ambiguous and there is even a function called `defaultMeshFilter`.

PR #113, opened 2017-05-02
Select option labels should be able to be falsey

GabeMedrash:
Currently, a `<Select />` option label will default to empty string if `option[labelKey]` is falsey. This makes it impossible to render a label with a '0', or 'false'. This fix changes behavior such that empty string will only be returned if `option[labelKey]` is undefined.

### DIFF
--- a/src/ui/select/src/test/select-option-label.test.js
+++ b/src/ui/select/src/test/select-option-label.test.js
@@ -26,10 +26,24 @@ describe('<SelectOptionLabel', () => {
         option: {
           name: 'Seattle',
         },
+        labelKey: 'other'
       };
       const wrapper = shallow(<FlatOptionLabel {...props} />);
 
       expect(wrapper).to.have.text('');
+    });
+
+    it('returns proper label if option[labelKey] is falsey but not undefined', () => {
+      const spec = [
+        { props: { option: { name: 0 }, labelKey: 'name' }, expectation: '0' },
+        { props: { option: { name: false }, labelKey: 'name' }, expectation: 'false' },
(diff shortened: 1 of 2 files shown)


src/ui/select/src/test/select-option-label.test.js
```diff
@@ -26,11 +26,25 @@ describe('<SelectOptionLabel', () => {
...
 
       expect(wrapper).to.have.text('');
     });
+
+    it('returns proper label if option[labelKey] is falsey but not undefined', () => {
+      const spec = [
+        { props: { option: { name: 0 }, labelKey: 'name' }, expectation: '0' },
+        { props: { option: { name: false }, labelKey: 'name' }, expectation: 'false' },
+        { props: { option: { name: null }, labelKey: 'name' }, expectation: 'null' },
+        { props: { option: { name: '' }, labelKey: 'name' }, expectation: '' },
```
aaronKeel: To be extra clear about falsy values, you may want to add test cases:
```
        { props: { option: { name: NaN }, labelKey: 'name' }, expectation: 'NaN' },
        { props: { option: { name: undefined }, labelKey: 'name' }, expectation: '' },
        { props: { option: { name: void(0) }, labelKey: 'name' }, expectation: '' },
```

aaronKeel:
I approve.

src/ui/select/src/test/select-option-label.test.js
GabeMedrash: Ah, thanks for the catch. I think adding the `NaN` test case is a good addition to this assertion block. I believe those others should be added to the assertion block above this one, which handles testing of undefined.

PR #114, opened 2017-05-05
disputed territories refactor

GabeMedrash:
This work effort captures changes necessary to support the breaking out of disputed territories into separate topojson objects.


src/utils/geo.js
```diff
@@ -26,24 +34,40 @@ export function extractGeoJSON(topology, layers) {
...
+ * @param {Object} objects
+ * @param {Array} order
+ * @return {{type: string, geometries: array}}
+ */
+export function concatTopoJSON(objects, order) {
+  return {
+    type: 'GeometryCollection',
+    geometries: reduce(order, (collection, layer) =>
+      concat(collection, objects[layer].geometries)
+    , []),
```
davschne: Style: I haven't seen the arrow function "concise" style (no braces, implicit return) used with a line break before. I think the usual block style would be a bit clearer in this context, particularly since there's another parameter following the function. Your call, though.

src/utils/test/geo.test.js
```diff
@@ -1,23 +1,43 @@
...
     { name: 'nonExistent' },
   ];
   const geo = getTopoJSON();
   const extractedGeoJSON = extractGeoJSON(geo, layers);
   const concatenatedGeoJSON = concatGeoJSON(extractedGeoJSON);
 
+  it('concatenates topojson objects together, in order', () => {
+    const expectedNumGeometries = reduce(geo.objects, (accum, object) =>
+      accum + object.geometries.length
+    , 0);
```
davschne: Same comment as above regarding arrow function body.

src/utils/geo.js
GabeMedrash: Good call, thanks. Airbnb now even has [a recommendation](https://github.com/airbnb/javascript#arrows--paren-wrap) regarding this, though it must be in a later version than what this lib is currently at.

src/utils/geo.js
```diff
@@ -16,7 +21,10 @@ const defaultMeshFilter = () => { return true; };
 export function extractGeoJSON(topology, layers) {
   return reduce(layers, (acc, layer) => {
     // make certain the layer exists on the topojson
-    if (!topology.objects.hasOwnProperty(layer.object)) return acc;
+    if (!layer.object
+      || (typeof layer.object === 'string' && !topology.objects.hasOwnProperty(layer.object))) {
```
GabeMedrash: Good point--I'll go with your first suggestion.

PR #115, opened 2017-05-10
bugfix/event handlers create new functions

### LINKED ISSUE #110: Shape component creates new functions every time it is rendered
May be better to pull this out of the `render` method.

GabeMedrash:
The way `eventHandleWrapper` was being used was creating new event handlers on every render for certain components. This work effort remedies that.

Closes #110.

### DIFF
--- a/src/ui/legend/src/legend-item.jsx
+++ b/src/ui/legend/src/legend-item.jsx
@@ -1,80 +1,110 @@
 import React, { PropTypes } from 'react';
 import classNames from 'classnames';
-import { propResolver, eventHandleWrapper } from '../../../utils';
+import bindAll from 'lodash/bindAll';
+
+import {
+  CommonDefaultProps,
+  propResolver,
+} from '../../../utils';
 
 import styles from './legend-item.css';
 import { Shape } from '../../shape';
 
-/**
- * label renderer
- */
-function renderLabel(props) {
-  /* eslint-disable react/prop-types */
-  const {
-    item,
-    LabelComponent,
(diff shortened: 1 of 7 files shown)


aaronKeel:
This is a great bug fix. An important lesson learned about being careful with function/object creation in our React components by reading this pr. Is `eventHandleWrapper` seeing its last days or should it be refactored? (open question)

GabeMedrash:
Thanks, @aaronKeel! Yeah, I was wondering about whether it's time to sunset `eventHandleWrapper`. Originally thought it would be a good way of enforcing consistency in event handler interfaces, but really, it's easy enough to get it right without a helper function that's probably doing more harm than good. I think v1 release will be a good target for extirpating it. At this point I believe the last trace of it is in `<Group />`.

src/ui/legend/src/legend-item.jsx
```diff
@@ -1,80 +1,110 @@
...
-        >
-          <path d="M-3,-3L3,3 M-3,3L3,-3" stroke="black" strokeWidth="1.5" />
-        </svg>
-      ) : null}
-      <div
-        className={classNames(styles['label-shape-wrapper'], {
-          [styles.clickable]: typeof onClick === 'function',
-        })}
-        onClick={onClick ? eventHandleWrapper(onClick, item) : null}
+export default class LegendItem extends React.Component {
```
aaronKeel: Since this was previously a functional component, would extending React.PureComponent be a better choice?

Issue #94, opened 2017-02-03
Scatter: support `symbolStyle` 


GabeMedrash:
`<MultiScatter` presumably allows you to pass in `props.symbolStyle`, but that isn't passed through `<Scatter />` to `<Symbol />`s.

GabeMedrash:
Closed by #120.

Issue #92, opened 2017-02-03
Shape: all of the `style` props should support functions


GabeMedrash:
Each is listed as a `CommonPropType.style`, which allows both objects and functions. If a function is passed, it should be called with current datum or whole data array, as applicable.

GabeMedrash:
Closed by #120.

PR #120, opened 2017-05-11
bugfix/scatter shape style

### LINKED ISSUE #91: Scatter: `style` prop inconsistently applied
It should be applied to the outermost, wrapping `<g>` within which `<Symbol />`s are rendered. `symbolStyle` should be exposed as a new prop and sent down into each `<Symbol />`.

GabeMedrash:
Closes #91, #92, and #94.

### DIFF
--- a/src/ui/shape/src/multi-scatter.jsx
+++ b/src/ui/shape/src/multi-scatter.jsx
@@ -23,6 +23,7 @@ export default class MultiScatter extends PureComponent {
     super(props);
 
     this.combineStyles = memoizeByLastCall(combineStyles);
+    this.castSelectionAsArray = memoizeByLastCall((selection) => castArray(selection));
   }
 
   render() {
@@ -89,7 +90,7 @@ export default class MultiScatter extends PureComponent {
                 data={scatterValues}
                 fill={color}
                 key={`scatter:${key}`}
-                selection={castArray(selection)}
+                selection={this.castSelectionAsArray(selection)}
                 style={scatterStyle}
                 shapeType={shapeType}
                 {...childProps}
--- a/src/ui/shape/src/multi-scatter.jsx
+++ b/src/ui/shape/src/multi-scatter.jsx
@@ -4,7 +4,14 @@ import { scaleLinear } from 'd3';
 import { castArray, map, pick } from 'lodash';
 import Scatter from './scatter';
 
-import { propResolver, PureComponent, CommonDefaultProps, CommonPropTypes } from '../../../utils';
+import {
+  combineStyles,
+  CommonDefaultProps,
+  CommonPropTypes,
+  memoizeByLastCall,
+  propResolver,
+  PureComponent,
+} from '../../../utils';
 
 /**
  * `import { MultiScatter } from 'ihme-ui'`
@@ -12,6 +19,12 @@ import { propResolver, PureComponent, CommonDefaultProps, CommonPropTypes } from
  * This is a convenience component intended to make it easier to render many `<Scatter />`s on a single chart.
  */
 export default class MultiScatter extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.combineStyles = memoizeByLastCall(combineStyles);
+  }
+
   render() {
     const {
       className,
@@ -20,11 +33,11 @@ export default class MultiScatter extends PureComponent {
       data,
       fieldAccessors,
       scatterClassName,
+      scatterStyle,
       scatterValuesIteratee,
       selection,
       style,
       shapeScale,
-      shapeStyle,
     } = this.props;
 
     const {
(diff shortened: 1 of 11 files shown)


src/ui/shape/test/scatter.test.js
```diff
@@ -50,6 +50,7 @@ describe('<Scatter />', () => {
       selectedClassName="selected"
       selectedStyle={{ stroke: 'aqua' }}
       shapeClassName="symbol"
+      shapeStyle={{ fill: 'bluesteel' }}
```
aaronKeel: 'bluesteel' or 'steelblue' Zoolander?

src/ui/shape/src/shape.jsx
```diff
@@ -287,16 +266,15 @@ Shape.propUpdates = {
...
-    return assign(accum, {
+    return assign({}, accum, {
       path: Shape.getPath(shapeType, nextProps.size),
       rotate: SYMBOL_ROTATE[rotate] || 0,
     });
   },
 
   // update style if datum, focused, focusedStyle, selected, selectedStyle, or style have changed
-  style: (accum, propName, prevProps, nextProps) => {
+  styles: (accum, propName, prevProps, nextProps) => {
```
aaronKeel: This is much better.

aaronKeel: I like this in combination with combineStyles. We are able to order the styles to be applied from left to right. Very nice.

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -50,13 +63,14 @@ export default class MultiScatter extends PureComponent {
...
       'shapeScale',
+      'shapeStyle',
     ]);
 
     return (
       <g
         className={className && classNames(className)}
         clipPath={clipPathId && `url(#${clipPathId})`}
-        style={style}
+        style={this.combineStyles(style)}
```
GabeMedrash: Yep, more or less. It accepts a list of items (or just a single item that it will cast as an array) that can either be inline style objects or a function. If a function, it is called along with any other args provided to `combineStyles`. Using `combineStyles` here mostly as a convenience for otherwise having to the check `if (typeof style === 'function') doSomething(); else somethingElse()`...just abstracting that away. Now that I look at it, I should definitely be more consistent with providing data as an argument in case `style` is a function. 

Seem like a good direction to you and a helpful utility? Too much?

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -75,8 +90,8 @@ export default class MultiScatter extends PureComponent {
                 data={scatterValues}
                 fill={color}
                 key={`scatter:${key}`}
-                selection={castArray(selection)}
-                style={shapeStyle}
+                selection={this.castSelectionAsArray(selection)}
+                style={scatterStyle}
```
GabeMedrash: Well, `combineStyles` is used internally inside `<Scatter />`, so my thinking was to just pass the `scatterStyle` prop down and let `Scatter` do its thing--would help avoid doing double work.

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -202,6 +217,11 @@ MultiScatter.propTypes = {
   scatterClassName: CommonPropTypes.className,
 
   /**
+   * inline styles applied to `<Scatter />`'s outermost wrapping `<g>`.
+   */
+  scatterStyle: CommonDefaultProps.style,
```
GabeMedrash: Nice catch, thanks

src/ui/shape/src/scatter.jsx
```diff
@@ -93,6 +99,7 @@ export default class Scatter extends PureComponent {
                 focused={focusedDatumKey === key}
                 selected={selectedDataMappedToKeys.hasOwnProperty(key)}
                 shapeType={resolvedShapeType}
+                style={shapeStyle}
```
GabeMedrash: Thinking again here to just pass it straight though as `Shape` will do its own combining of styles. Make sense?

src/ui/shape/src/shape.jsx
GabeMedrash: Thanks!

src/ui/shape/test/scatter.test.js
GabeMedrash: ha, whoops

GabeMedrash:
Thanks for the great review! Good question about whether we would ever *not* want `combineStyles` to be memoized. Probably always a good thing,  but the strategy used here (`memoizeByLastCall`) makes it a little funky to just incorporate that as an implementation detail. I suppose we could do something like:
```
function memoizedCombineStyles() {
  return memoizeByLastCall(combineStyles);
}
```

Then it would be used like:
```
class Foo {
  constructor(props) {
    super(props);
    this.combineStyles = memoizedCombineStyles(); // note function call
  }

  ...
}
```

What do you think?

PR #121, opened 2017-05-12
Legend: standardize className and style props

### LINKED ISSUE #88: Legend: change prop names for consistency
`wrapperClassName` -> `className`
`wrapperStyles` -> `style`

Will break mortality, gbd-search, potentially PSS

GabeMedrash:
Closes #88.

### DIFF
--- a/src/ui/legend/src/legend.jsx
+++ b/src/ui/legend/src/legend.jsx
@@ -55,15 +55,16 @@ export default class Legend extends React.Component {
 
   render() {
     const {
-      wrapperClassName,
-      wrapperStyles,
-      ulClassName
+      className,
+      listClassName,
+      listStyle,
+      style,
     } = this.props;
 
     return (
-      <div className={classNames(styles.container, wrapperClassName)} style={wrapperStyles}>
+      <div className={classNames(styles.container, className)} style={style}>
         {this.renderTitle()}
-        <ul className={classNames(styles.list, ulClassName)}>
+        <ul className={classNames(styles.list, listClassName)} style={listStyle}>
           {this.renderItemList()}
         </ul>
       </div>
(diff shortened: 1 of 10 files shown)


src/ui/legend/src/legend.jsx
```diff
@@ -112,6 +119,16 @@ Legend.propTypes = {
...
 
   /**
+   * className applied to `<ul>`, which wraps legend items
+   */
+  listClassName: CommonPropTypes.className,
+
+  /**
+   * inline styles applied to `<ul>`, which wraps legend items
+   */
+  listStyle: PropTypes.object,
```
aaronKeel: Use CommonPropTypes.style, or is a function not valid?

src/ui/legend/src/legend.jsx
```diff
@@ -150,6 +167,11 @@ Legend.propTypes = {
   ]).isRequired,
 
   /**
+   * inline styles applied to outermost, wrapper `<div>`
+   */
+  style: PropTypes.object,
```
aaronKeel: Use CommonPropTypes.style, or is a function not valid?

src/ui/legend/src/legend.jsx
GabeMedrash: For this and your comment below, I wasn't entirely certain what to do, but you're probably correct in that I should make them accept functions. The elements in question (basically just wrappers) aren't themselves "data-driven," which is why I thought it might be overkill to make this accept a function that accepts data as an argument. But in idiomatic d3, you're still able to pass a function to `selection.style(d => ...)` in cases like this. The `d` that's passed in is just the whole data structure passed into the bind. So I think what I'll is have this as well as `props.style` accept a function, and the first argument will be `props.items`.

PR #130, opened 2017-08-02
Dynamic axis ticks

GabeMedrash:
`<XAxis />` and `<YAxis />` are now composed from an `orientAxis` HOC, which provides each with functionality for automatically filtering out tick values as available space (width or height) shrinks. To enable, pass `<XAxis />`/`<YAxis />` `autoFilterTickValues` (boolean prop).


src/ui/axis/src/utils.js
```diff
@@ -77,3 +88,28 @@ export function calcLabelPosition(orientation, translate, padding, center) {
...
+
+export function filterTickValuesByWidth(ticks, {
+  tickFontSize,
+  tickFontFamily,
+  tickFormat,
+  width
+}) {
+  const canvasContext = window && window.document.createElement('canvas').getContext('2d');
+
+  const widestTickLabel = reduce(map(ticks, tickFormat), (widest, tick) =>
```
aaronKeel: Could `canvasContext` be avoided by approximating string width with `string.length * tickFontSize`, with the assumption that generally character glyphs are not wider than the font size. This would be an over estimate, and your canvas solution is great, but a simpler solution might be a better choice.

src/ui/axis/src/x-axis.jsx
```diff
@@ -11,63 +12,13 @@ import Axis, { AXIS_SCALE_PROP_TYPES } from './axis';
...
-  /**
-   *  scales are provided by axis-chart, only x scale is used by XAxis
-   */
-  scales: atLeastOneOfProp(X_AXIS_SCALE_PROP_TYPES),
-};
-
-XAxis.defaultProps = {
+XAxis.defaultProps = assign({}, XAxis.defaultProps, {
   orientation: 'bottom',
   scales: { x: scaleLinear() },
```
aaronKeel: Provide just `scale: scaleLinear()`?

src/ui/axis/src/y-axis.jsx
```diff
@@ -11,63 +12,13 @@ import Axis, { AXIS_SCALE_PROP_TYPES } from './axis';
...
-  /**
-   *  scales are provided by axis-chart, only y scale is used by YAxis
-   */
-  scales: atLeastOneOfProp(Y_AXIS_SCALE_PROP_TYPES),
-};
-
-YAxis.defaultProps = {
+YAxis.defaultProps = assign({}, YAxis.defaultProps, {
   orientation: 'left',
   scales: { y: scaleLinear() },
```
aaronKeel: Provide just `scale: scaleLinear()`?

src/ui/axis/src/orientedAxis.js
```diff
@@ -0,0 +1,143 @@
...
+};
+
+/**
+ *
+ * @param {React.Component} AxisComponent
+ * @param {string} orientation - oneOf([Orientation.HORIZONTAL, Orientation.VERTICAL])
+ * @return {OrientedAxis}
+ */
+export default function orientAxis(AxisComponent, orientation) {
+  class OrientedAxis extends PureComponent {
```
GabeMedrash: This is the old composition vs. inheritance debate. The React community seems to me to have [fallen squarely on the composition side](https://facebook.github.io/react/docs/composition-vs-inheritance.html) of the debate. Not to say that's a reason in and of itself, but I think there's something to be said to keeping style consistent with the community you work within. But I do actually believe this is a perfectly [reasonable use of inheritance](https://www.thoughtworks.com/insights/blog/composition-vs-inheritance-how-choose). Ultimately, I think we should use an HoC because it gets rid of some of the wonkiness we had to include with PropTypes checking to account for the orientation of the axis.

src/ui/axis/src/orientedAxis.js
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * required if and only if `props.scales` is not provided
+     */
+    scale: PropTypes.func,
+
+    /**
+     * required if and only if `props.scale` is not provided
+     */
+    scales: PropTypes.shape({
```
GabeMedrash: This is an unfortunate consequence of an implementation detail of `<AxisChart />`. That component passes `scales` down to each of its children. So when `<Axis />` is a child of `<AxisChart />`, it must accept `scales`. Ultimately, I think we should think through the design of `<AxisChart />` a little differently, maybe such that it compose it's own axes, and it will know, internally, which axis to pass an x-scale, and which to pass a y-scale.

src/utils/array.js
```diff
@@ -28,3 +22,18 @@ export function linspace(domain, length) {
...
   });
 }
+
+/**
+ * filter list of elements to length equal to or less than maxLength
+ * @param {array} list
+ * @param {number} maxLength
+ * @return {array}
+ */
+export function choose(list, maxLength) {
```
GabeMedrash: Hm. Not sure I agree with you regarding the usual meaning of "choose."  But I'm happy to rethink the name of this utility. I'd be reticent to use "partition," if only because [lodash has a function named partition](https://lodash.com/docs/4.17.4#partition) that does something quite different

src/ui/axis/src/utils.js
GabeMedrash: That's a good question for me to investigate. From what I've read, font-size is more closely related to the rendered height (height of the em square, to be specific) of the font than it's width, but I might do some experiments to see if we can avoid actually rendering the string.

src/ui/axis/src/x-axis.jsx
GabeMedrash: Good call

src/ui/axis/src/orientedAxis.js
aaronKeel: Thanks for the explanation. The react page provides great context and now I see that composition is more consistent with the library.

src/utils/array.js
aaronKeel: I was just trying to lean towards [partitioning an interval](https://en.wikipedia.org/wiki/Partition_of_an_interval). Another idea is [taking a spread](https://stackoverflow.com/questions/9873626/choose-m-evenly-spaced-elements-from-a-sequence-of-length-n). I guess my main concern is that 'choose' by itself is ambiguous about what it is choosing.

PR #133, opened 2017-10-16
Update react, react-dom, react-addons-* to 15.6

davschne:
Also adding react-test-renderer, needed to run our test suite with this version of React.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ node_js: stable
 
 os: osx
 
-osx_image: xcode8
+osx_image: xcode7.3
 
 notifications:
   email: false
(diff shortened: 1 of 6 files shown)


kbeame:
@GabeMedrash Thanks for the ideas- it still seems to be the same problem however, origin unknown.

GabeMedrash:
Hmmm...what if we just skipped straight to 16.x? That way we could full-on update all test dependencies (Enzyme has a new way to set itself up for React 16, etc.). Wild guess, but I have a hunch that whatever we'd have to do to get it working in 15.6 would also make it work with 16.  I don't think we'd have to change anything in `src` to accomplish this...

PR #135, opened 2017-12-08
Incremental react verion to 15 6

kbeame:
This includes all of the updates that I made to bring ihme-ui over to use React v16.
However because of the issues with ihme-react-select's compatibility with React v16 I have decided to move ahead with updating ihme-ui to React v15.6 as a temporary measure.
Currently ihme-react-select cannot use React v16 as its necessary react peer dependency.

The benefits of this is that I have already done the updates needed to move over to React v16 (once we update ihme-react-select).
TODO for eventually migrating to React v16:
- Update dependency "ihme-react-select"
- Update dependency "react" and "react-dom" in ihme-ui to v16
- Get rid of dependency "react-test-renderer"
- Change dependency "enzyme-adapter-react-15" to "enzyme-adapter-react-16" (see setup-jsdom.js for single import update)


src/ui/compositions/map/src/map.jsx
```diff
@@ -102,7 +103,7 @@ export default class Map extends React.Component {
     // only update if data is not being currently loaded,
     // and when props have changed,
     return !nextProps.loading
-            && PureRenderMixin.shouldComponentUpdate.call(this, nextProps, nextState);
+            && shouldPureComponentUpdate.call(this, nextProps, nextState);
```
davschne: I believe this should be `shouldPureComponentUpdate(this.props, this.state, nextProps, nextState)`. Note that the function signature is different from `PureRenderMixin.shouldComponentUpdate`. `.call` is unnecessary, I think, since the function doesn't use `this` context.

src/ui/axis-chart/src/axis-chart.jsx
```diff
@@ -62,7 +68,7 @@ export default class AxisChart extends React.Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
-    return !nextProps.loading && PureRenderMixin.shouldComponentUpdate(this, nextProps, nextState);
+    return !nextProps.loading && shouldPureComponentUpdate(this, nextProps, nextState);
```
davschne: I believe this should be `shouldPureComponentUpdate(this.props, this.state, nextProps, nextState)`. Note that the function signature is different from `PureRenderMixin.shouldComponentUpdate`.

package.json
```diff
@@ -92,11 +94,10 @@
...
     "lodash": "~4.5",
-    "react": "~15.1",
-    "react-addons-pure-render-mixin": "~15.1",
-    "react-addons-shallow-compare": "~15.1",
-    "react-dom": "~15.1",
-    "react-virtualized": "~7.24",
+    "prop-types": "^15.6.0",
+    "react": "~15.6",
+    "react-dom": "~15.6",
+    "react-virtualized": "^9.00",
```
davschne: I suggest using the tilde (~) for the new package dependencies. I think that's a safer practice, because it will lock down the packages to the specified minor version (patch updates will still be applied).

src/utils/react.js
```diff
@@ -1,9 +1,18 @@
...
-import PureRenderMixin from 'react-addons-pure-render-mixin';
+import { shallowEqual } from './objects';
 
-export class PureComponent extends React.Component {
-  constructor(props) {
-    super(props);
-    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate.bind(this);
-  }
+// Export PureComponent for backwards compatibility (introduced before React.PureComponent existed).
+// export { PureComponent } from 'react';
```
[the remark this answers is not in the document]
kbeame: Thanks @davschne - I made all of your requested changes with the exception of continuing to export PureComponent. I'm willing to do that too but want more people to weigh in so I'm holding off for now.

src/ui/compositions/map/test/map.test.js
```diff
@@ -18,12 +18,13 @@ describe('<Map />', () => {
       ];
 
       const keysOfVisibleGeometries = [1, 2];
+      console.log(filterData(data, keysOfVisibleGeometries, 'key'));
```
aaronKeel: remove console log

davschne:
Uh oh, failing tests in the Travis build.

src/ui/choropleth/test/feature-layer.test.js
```diff
@@ -17,10 +17,11 @@ describe('Choropleth <FeatureLayer />', () => {
   const features = getGeoJSON('states', 'feature').features;
 
   const data = getLocationIds(features).reduce((accum, locationId) => {
+    const meanValue = (Math.floor(Math.random() * 100));
```
aaronKeel:
```javascript
Math.floor(Math.random() * 100 + 1)
```
will solve this.

PR #138, opened 2018-02-09
Upgrade react to version 16 and refactor select component

aaronKeel:


src/ui/select/src/select.jsx
```diff
@@ -24,67 +27,53 @@ export default class Select extends React.PureComponent {
...
+        multi={multi}
+        optionHeight={optionHeight}
+        optionRenderer={optionRenderer({ hierarchical, multi })}
+        placeholder={!multi && (placeholder || 'Add/Remove...')}
+        removeSelected={false}
+        resetValue={resetValue || (multi ? [] : null)}
         searchable
-        valueComponent={Value}
-        valueRenderer={multiValueRenderer}
+        valueComponent={multi && retNull}
```
davschne: I've been looking at the documentation for `react-select`, but I'm having a hard time figuring out the purpose of this `valueComponent` prop. It does seem that it's required when `multi` is `true` (https://github.com/JedWatson/react-select/blob/c9d027387a77b3695fdfbb3d4de25ffcea9505ee/src/Select.js#L818-L835). Am I correct in thinking that we don't really have any use for this prop, but that we have to pass something to prevent an exception when `multi` is `true`? I also wonder if we'd ever have to worry about this other condition in which the `valueComponent` would be rendered: https://github.com/JedWatson/react-select/blob/c9d027387a77b3695fdfbb3d4de25ffcea9505ee/src/Select.js#L836-L850.

src/ui/select/src/test/select.test.js
```diff
@@ -0,0 +1,67 @@
...
+    const wrapper = mount(
+      <Select
+        {...baseProps}
+        value={null}
+      />
+    );
+
+    expect(wrapper.find('.Select-placeholder')).to.have.length(1);
+    expect(wrapper.find('.Select-placeholder')
+      .at(0).html()).to.equal('<div class="Select-placeholder">Add/Remove...</div>');
```
davschne: Two thoughts here:
(1) I think it would make maintenance easier if you stored default values (like 'Add/Remove...' here) somewhere they could be imported by this test file. They could remain in `select.jsx` or they could go in a separate file from which both `select.jsx` and this file could import them. That way, if we ever decided to change a default, it would only have to be changed in one place.
(2) I'm concerned that comparing on the HTML string might be a bit fragile. It seems like this could break under a variety of circumstances (like adding more attributes to the `<div>`, changing the `<div>` to some other kind of element, adding more classes to the element, etc.). If you're mainly concerned about the placeholder text in this test, wouldn't it be simpler just to test the contents of the element?

src/ui/select/src/test/select.test.js
```diff
@@ -0,0 +1,67 @@
...
+    const wrapper = mount(
+      <Select
+        {...baseProps}
+        value={[]}
+        multi
+      />
+    );
+
+    expect(wrapper.find('input')).to.have.length(1);
+    expect(wrapper.find('input').at(0).props().placeholder).to.equal('Add/Remove... (0)');
```
davschne: Same comment about default values.

package.json
```diff
@@ -65,38 +66,38 @@
...
-    "style-loader": "~0.17",
-    "webpack": "~2.2"
+    "style-loader": "^0.20.1",
+    "webpack": "^3.10.0"
   },
   "dependencies": {
     "babel-runtime": "~6.6",
     "classnames": "~2.2",
-    "d3": "~4.2",
-    "ihme-react-select": "git+https://github.com/ihmeuw/ihme-react-select.git",
```
davschne: Cool; I'm hoping that removing this dependency will mean projects using IHME-UI will no longer need to have Git installed.

src/ui/select/src/select.jsx
aaronKeel: For the first case, when multi is true, react-select wants to render "tag" or value components for each selection in a multi-select as seen in the demo/example page: (http://jedwatson.github.io/react-select/). Since that is not a desired feature for our library, I had to work around that value component part of the code and also maintain the positioning of the input component where the user can type and search the options. It is a bit of complexity that is not ideal, but in my opinion it beats out the alternatives of trying to create a new component or forking and refactoring the component again.

aaronKeel: For the second case, when multi is false, the component uses the default setting in line 1312.

src/ui/select/src/test/select.test.js
aaronKeel: (2) Yes, I was hoping to test just the placeholder text and not the div element. I'll look at the enzyme api to find a way to do that.

PR #140, opened 2018-02-28
Option renderer plus styles

aaronKeel:
- This PR updates the demo app for the select component with some of the updated features below.

- `optionRenderer` take in an `optionStyle` prop.

- Remove blue focused select box.

- Optional `clearable` prop.

- Optional `optionRenderer` to be passer as `innerRowRenderer` in React-Virtualized-Select.


aaronKeel:
run `npm run demo select` to view the select demo.

src/ui/select/demo/app.jsx
```diff
@@ -1,40 +1,133 @@
...
 
-const randomizer = randomUniform(0, 4);
+const random = Math.random;
+
+function* hierarchyGenerator() {
+  let level = 1;
+  yield level;
+  while (true) {
+    const q = random();
+    if (level === 1) {
```
davschne: I wonder if it would work to combine one or more of these conditions, like
```javascript
if (level === 1 && q < 0.5)
```
Behavior would be different for a few cases; wonder if that matters.

src/ui/select/demo/app.jsx
```diff
@@ -1,40 +1,133 @@
...
 
-import { Select } from '../';
+import { Select, SelectOption } from '../';
 import { default as cities } from './cities';
 import style from './app.css';
 
-const randomizer = randomUniform(0, 4);
+const random = Math.random;
+
+function* hierarchyGenerator() {
```
davschne: Generator functions for the win!!

src/ui/select/demo/app.jsx
```diff
@@ -1,40 +1,133 @@
...
+    }
+
+    if (option[valueKey] === 'select all') {
+      return (
+        <div
+          key={key}
+          style={style}
+          title={option.title}
+        >
+          <SelectOption
```
davschne: I'm struggling to understand why you don't pass props `hierarchical` and `multi` in this case.

src/ui/select/demo/app.jsx
```diff
@@ -1,40 +1,133 @@
...
+    }
+    yield level;
+  }
+}
+
+const levelG = hierarchyGenerator();
 
 const hierarchicalCities = cities.map((city) => {
-  const level = Math.floor(randomizer());
+  const {value: level} = levelG.next();
```
jimvermillion: What's the style rule for spaces next to curly braces? I feel like it'd be more like with the precedence I've seen in the codebase before now.

> const { value: level } = levelG.next();

src/ui/select/demo/app.jsx
```diff
@@ -1,40 +1,133 @@
...
+  const colors = [
+    '#26c',
+    '#C13',
+    '#c71',
+    '#b1c',
+    '#eee',
+  ];
+
+  const color = colors[Math.floor(random() * 10)];
+  if (color) return { color };
```
jimvermillion: Small nit: I'd prefer to see this return in one line.

> return color ? { color } : {};

src/ui/select/src/select.jsx
```diff
@@ -76,12 +79,24 @@ export default class Select extends React.PureComponent {
 }
 
 const selectPropTypes = {
+  /* should it be possible to reset value */
```
davschne: I'd make this a sentence and put a question mark at the end.

src/ui/select/demo/app.jsx
aaronKeel: Weird it didn't give a lint err.

src/ui/select/demo/app.jsx
aaronKeel: I want the condition chain to break if level is 1 or 4.

src/ui/select/demo/app.jsx
aaronKeel: Without `hierarchical` the option will not be indented. Without `multi` the option will not have an input checkbox. That is all.

davschne: Oh, okay, I think I've got it now: you don't want 'select all' to be indented or have an input checkbox.

PR #142, opened 2018-03-07
React move scatter

jimvermillion:
I've implemented basic animations in the scatter component and am looking for feedback in the patterns I've developed before I continue this effort. 

The implementation attempts to be as dry as possible while still rendering quite different components when `animate` is either true or false. This approach seems has pros of consolidating any future efforts of changes that would go on the animated as well non-animated scatter, which is nice. If the implementations were completely separate I could see instances of forgetting to make changes on one or the other. Its cons are that the implementation could be harder to read. Please let me know if it is understandable, or potentially able to be improved, or if I need to take a less cohesive approach.

Probably the biggest difference I had to make in the non-animated scatter implementation was to make it conform to the shape react-move requires so they could utilize the same processing functions (Fn `processDataSet`). Hopefully that seems logical.

Small changes were necessary in the Shape component to accommodate react-move using values wrapped in arrays to distinguish what should be animated and what should not be animated (I wish they hadn't made that choice -- look at the `Transitions` portion of the [react-move documentation](https://react-move.js.org/#/documentation/node-group)).

If this all seems well and good, my next steps are:
- Add tests for the animation functionality in scatter.
- Add a scatter component to the new demo app in the project's root.
- Continue this pattern to add animation to `Line`, `MultiLine`, `Axis` (for ticks, labels), and anything else you think should be animated in IHME-UI.

Thanks!
-Jim


jimvermillion:
@aaronKeel, it looks like the last `react-move` commit `8a95d9f` didn't have a coverage report. I don't know the ins and outs of coverage, but since `react-move` won't get merged into master for a while (and subsequent commits should have coverage) I think this should be okay. Thoughts?

src/ui/shape/src/scatter.jsx
```diff
@@ -28,92 +31,157 @@ import Shape from './shape';
...
+    const translateX = Scatter.getCoordinate(
+      propResolver(datum, dataAccessors.x),
+      scales && scales.x,
+    );
+    const translateY = Scatter.getCoordinate(
+      propResolver(datum, dataAccessors.y),
+      scales && scales.y,
+    );
+
+    // *react-move requires that values be wrapped in an array if they are to be animated.
```
Ryshackleton: It seems like you're having to do a lot of case testing to handle this array-ification of props when animating.  Would it be worth making default behavior in IHME-UI be to wrap props in arrays to avoid this?  I was thinking that if all of the components start to get animation, there could be a lot of extra case-checking throughout.   I don't know enough about the underlying implementation of IHME's normal prop handling to know if that's an absurd idea or not, but I thought I'd throw it out there.

jimvermillion: Hmmm, it's worth thinking about. But I don't think there is an awesome way around it. 

If we wrap everything, and someone wants a certain property to not animate (say color) while making another property animate (say position) react-move will see the color being wrapped and animate it. That's how it knows what to animate using the node-group [scroll to the "Tranisitions" portion](https://react-move.js.org/#/documentation/node-group). It's not something I've been super happy about working around, but I also don't find it too much of an obstacle if it can get our charts moving.

Ryshackleton: Oh, I didn't realize that it was using the array wrapping as a flag for when to animate.  I thought that was just a convention.

src/ui/shape/src/multi-scatter.jsx
```diff
@@ -104,6 +109,11 @@ export default class MultiScatter extends React.PureComponent {
 }
 
 MultiScatter.propTypes = {
+  /**
+   * Whether to animate the multi-scatter component or not.
+   */
+  animate: PropTypes.boolean,
```
aaronKeel: A couple thoughts about props related to animate.
1) Suppose animate were an object
```javascript
{
  enter() {...},
  leave() {...},
  start() {...},
  update() {...}
}```
and animate if `animate` is truthy.
2) Make `shouldAnimate` a boolean, and include `animate` as the object described above.

Either way, we reduce the number of props that depend on `animate` being true, if it is false `enter`, `leave`, `start` and `update` don't matter.

Comment (jimvermillion) [reply src/ui/shape/src/multi-scatter.jsx]: @aaronKeel, are these thoughts just for multi-scatter?

For `shouldAnimate`, would that be a method, like `this.shouldAnimate` or another prop?
I am imagining you wanting to follow the [formidable approach](https://github.com/FormidableLabs/victory-chart/blob/master/src/components/victory-scatter/victory-scatter.js#L64). I don't necessarily see the need for the method `shouldAnimate` currently, seeing as Formidable put that in to handle react-native implementations. But if it makes it easier to understand, I'm happy to include it.

I am seeing merit in nesting `enter`, `leave`, `start`, `update` within the `animate` prop though. It makes those methods more intuitive. It does kind of go away from keeping props as flat as possible, which is something that Brett and Gabe always attempted to do...just something to consider.

What do you think about `animate` being `boolean | <the shape you described above>`?

Comment (aaronKeel) [reply src/ui/shape/src/multi-scatter.jsx]: formidable must have been where I saw that. You are right about the props being flat though. Flat props are better.

File: src/ui/shape/src/multi-scatter.jsx [hunk 3]
```diff
@@ -110,9 +106,59 @@ export default class MultiScatter extends React.PureComponent {
 
 MultiScatter.propTypes = {
   /**
-   * Whether to animate the multi-scatter component or not.
+   * Whether to animate the scatter component (using default `start`, `update` functions).
+   * Optionally, an object that provides functions that dictate behavior of animations.
    */
-  animate: PropTypes.boolean,
+  animate: PropTypes.oneOfType([
```
aaronKeel: I think it’s probably ok since they are props specific to the component being animated. Now since `animate` is an object, should there be defaults for `enter, leave, start, and update`? What if I pass and object and `animate` but don't provide methods?

jimvermillion: If you gave an empty object to `animate` -> `shouldAnimate` would evaluate to `true`, yielding the same effect as providing `animate` with the boolean of `true`.

While those specific properties on the `animate` prop do not have defaults, [there is default functionality](https://github.com/ihmeuw/ihme-ui/blob/react-move-scatter/src/ui/shape/src/scatter.jsx#L162) with regards to how we treat `start` and `enter`. They just evaluate the datum like they normally would and `react-move` provides the default `timing` properties.

jimvermillion: An adjacent concern I have has to do with what `react-move` expects when it's receiving what is returned from its animation functions.
ie, it needs to know the state of the next position, color, etc to animate it. In that same object, react-move expects the transition information. So we're forced to couple the datum computation with how we would like the component to animate!

I'm thinking of only allowing the users of the `<Scatter />` component choose the animations to override, rather than the data computations itself. This would drastically lower the bar of what we'd expect the user to know for them to supply custom transitions/interrupt events. However, the tradeoff would be that they would be much more limited on extensive transitions. For instance, if someone wanted a different duration for their color-fill than their translation that wouldn't be possible. But it'd be the price of a more easily consumed component. (Otherwise they can make custom charts using `react-move` and our `Shape` components themselves). Any thoughts on that?

src/ui/shape/src/shape.jsx
```diff
@@ -144,7 +145,11 @@ Shape.propTypes = {
   /**
    * Fill color for path.
    */
-  fill: PropTypes.string,
+  fill: PropTypes.oneOfType([
```
davschne: `fill` could be an array of `number`s but not a `number`? What would the number represent?

src/utils/props.js
```diff
@@ -36,6 +36,85 @@ export const CommonDefaultProps = {
   identity,
 };
 
+/**
+ * Event functions that happen during transition. The given functions are passed no arguments but
+ * have access to the context to which they were created.
```
davschne: "...context _in_ which they were created."

src/utils/props.js
```diff
@@ -36,6 +36,85 @@ export const CommonDefaultProps = {
...
+ * A function that returns an object describing how the state should initially start.
+ * The function is passed:
+ *   `value`: number | string - computed by component
+ *   `rawDatum`: {}           - datum used to compute `value`.
+ *   `index`: number          - index of the `rawDatum`.
+ *
+ * signature: (value, rawDatum, index) => {
+ *   [keyof Prop]: PropTypes.oneOfType([Proptypes.string, PropTypes.number]);
+ * }
+ * @type {AnimateMethod}
```
davschne: Does this line belong here?

src/utils/props.js
```diff
@@ -36,6 +36,85 @@ export const CommonDefaultProps = {
...
+ * @type {AnimateStart}
+ */
+export const AnimateStart = PropTypes.func;
+
+/**
+ * A function that returns an object describing how the state should transform.
```
davschne: Have you considered defining this and other function signatures using JSDoc callback declarations like those at the bottom of this file? I imagine this would be helpful for generating documentation.

src/utils/props.js
```diff
@@ -36,6 +36,85 @@ export const CommonDefaultProps = {
...
+ */
+export const AnimateTiming = PropTypes.shape({
+  delay: PropTypes.number,
+  duration: PropTypes.number,
+  ease: PropTypes.func,
+});
+
+/**
+ * Values not wrapped in an array will not be animated.
+ * [detailed in react-move](https://react-move.js.org/#/documentation/node-group).
```
davschne: No `@type {AnimatableValue}` declaration?

src/utils/props.js
```diff
@@ -36,6 +36,85 @@ export const CommonDefaultProps = {
...
+ * signature: (value, rawDatum, index) => {
+ *   [keyof Prop]: AnimatableValue;
+ *   events: AnimateEvents;
+ *   timing: AnimateTiming;
+ * }
+ * @type {AnimateMethod}
+ */
+export const AnimateMethod = PropTypes.func;
+
+export const AnimateProp = PropTypes.shape({
```
davschne: No JSDoc for this one?

src/ui/shape/src/scatter.jsx
```diff
@@ -1,18 +1,22 @@
...
+import findIndex from 'lodash/findIndex';
+import isFinite from 'lodash/isFinite';
+import keyBy from 'lodash/keyBy';
+import map from 'lodash/map';
+import pick from 'lodash/pick';
+import sortBy from 'lodash/sortBy';
 
 import {
+  animationProcessorFactory,
+  animationStartFactory,
```
davschne: `eslint` tells me `animationStartFactory` is defined but never used.

src/ui/shape/src/scatter.jsx
```diff
@@ -28,92 +32,188 @@ import Shape from './shape';
...
+      fill: processedFill,
+      shapeType: resolvedShapeType,
+      translateX,
+      translateY,
+    };
+  }
+
+  processDataSet(data) {
+    const { key } = this.props.dataAccessors;
+    return data.reduce((accum, datum) => {
```
davschne: Why `reduce`? Couldn't this be done as a `map` operation? Complexity here is quadratic in the length of the input array, since by spreading `accum` below, you're recopying all prior array elements on each iteration. If you `map` instead, complexity will be linear.

src/ui/shape/src/scatter.jsx
```diff
@@ -28,92 +32,188 @@ import Shape from './shape';
...
+        {this.renderScatter}
+      </NodeGroup>
+    );
+  }
+
+  shouldAnimate() {
+    return !!this.props.animate;
+  }
+
+  render() {
```
davschne: Very elegant `render`; love it!

src/ui/shape/src/scatter.jsx
davschne: And this seems to be what's causing the build to fail.

src/ui/shape/src/scatter.jsx
```diff
@@ -257,38 +351,89 @@ Scatter.propTypes = {
...
+  shapeStyle: undefined,
   shapeType: 'circle',
+  start: undefined,
+  update: undefined,
 };
 
 Scatter.propUpdates = {
+  animationProcessor: (state, _, prevProps, nextProps) => {
+    if (!propsChanged(prevProps, nextProps, ['animate'])) return state;
+    const animationProcessor = animationProcessorFactory.bind(
```
aaronKeel: Could use lodash `partial`.

src/ui/shape/src/scatter.jsx
```diff
@@ -28,92 +31,183 @@ import Shape from './shape';
...
+    } = this.state;
+
+    return (
+      <Shape
+        className={shapeClassName}
+        key={key}
+        datum={data}
+        fill={fill}
+        focused={focusedDatumKey === key}
+        selected={selectedDataMappedToKeys.hasOwnProperty(key)}
```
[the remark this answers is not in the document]
jimvermillion: Cool, I will update since `has` is more semantic/brief.

src/ui/shape/src/scatter.jsx
```diff
@@ -28,92 +31,183 @@ import Shape from './shape';
...
+      shapeType: resolvedShapeType,
+      translateX,
+      translateY,
+    };
+  }
+
+  processDataSet(data) {
+    const { key } = this.props.dataAccessors;
+    // This data shape is compatible with animated and non-animated components.
+    return data.map(
```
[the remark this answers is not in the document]
jimvermillion: Cool, I think I'm going to update the use of `_.map` to

src/ui/shape/src/scatter.jsx
jimvermillion: Sounds more semantic to use `partial` so I'll update.

PR #143, opened 2018-03-15
Tooltip fix

aaronKeel:
This PR is a fix to a bug in firefox running on windows. It gets rid of render artifacts when the tooltip is set to `visibility:hidden`. According to [this article from 2012](https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/) translate has better performance. But maybe times have changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ihme-ui",
   "description": "Visualization tools from the Institute for Health Metrics and Evaluation",
-  "version": "0.29.3",
+  "version": "0.29.4",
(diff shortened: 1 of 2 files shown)


jimvermillion:
Looks good! Thanks for taking care of that.

aaronKeel:
@davschne 
I tested the tooltip demo on browsers.
Windows: Firefox, Edge, IE, Chrome
OSX: Firefox, Safari, Chrome.
Are we supporting any others?
I didn't go into older versions of these browsers.

PR #144, opened 2018-03-17
Bugfix/react move multi scatter

jimvermillion:
Fix nested data structure issue.

- make processDatum static that has props passed to it.
- pass into animation function.
- I'm not in love with the pattern, I'm open to alternatives.

### DIFF
--- a/src/ui/shape/src/scatter.jsx
+++ b/src/ui/shape/src/scatter.jsx
@@ -174,13 +174,13 @@ export default class Scatter extends React.PureComponent {
   }
 
   render() {
-    const data = this.processDataSet(this.state.sortedData);
+    if (this.shouldAnimate()) {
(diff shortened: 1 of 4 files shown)


jimvermillion:
Note: I will have to update all animation tests to match the new processing pattern.
That's why travis is unhappy.

src/ui/shape/src/scatter.jsx
```diff
@@ -177,7 +178,7 @@ export default class Scatter extends React.PureComponent {
```
aaronKeel:
```
if (!this.shouldAnimate()) {
     const data = this.processDataSet(this.state.sortedData);
     return this.renderScatter(data)
}
return this.renderAnimatedScatter(this.state.sortedData);
```
will avoid unnecessary `processDataSet`.

aaronKeel:
Looks good!

PR #147, opened 2018-05-24
Bug fix - Fix styles for non-hierarchical options

komalali:
The options don't bold as expected for non-hierarchical dropdowns because the styles are not being transferred to the DOM node.

### DIFF
--- a/src/ui/select/src/select-option-label.jsx
+++ b/src/ui/select/src/select-option-label.jsx
@@ -8,7 +8,9 @@ import { default as getValue } from 'lodash/get';
  */
 const flatOptionPropTypes = {
   // option object
-  option: PropTypes.object.isRequired,
+  option: PropTypes.shape({
+    bold: PropTypes.bool,
+  }),
 
   // key on option that holds its label
   labelKey: PropTypes.string.isRequired,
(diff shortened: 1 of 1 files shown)


davschne:
Uh oh, looks like the `npm install` is failing because of a compilation error in `node_gyp`. Nothing to do with the PR, I'm sure. How annoying!

src/ui/select/src/select-option-label.jsx
```diff
@@ -15,7 +15,15 @@ const flatOptionPropTypes = {
 };
 
 export function FlatOptionLabel(props) {
-  return <span>{`${getValue(props.option, [props.labelKey], '')}`}</span>;
+  return (
+    <span
+      style={{
+        fontWeight: props.option.bold ? 'bold' : 'normal',
```
jimvermillion: The `flatOptionPropTypes` property for `option` is just an object with no shape. I think adding a shape to it that includes bold would make sense in this PR. ie, something like the shape shown for the `hierarchicalOptionPropTypes` but without `level`:

```
  option: PropTypes.shape({
    bold: PropTypes.bool,
  }),
```

PR #156, opened 2018-06-04
React move

jimvermillion:
This PR would allow for animation in the Line, MultiLine, Scatter, MultiScatter, and Area components in the ./src/ui/shape/ directory. 

The `react-move` branch has been my main working branch for this effort and it includes the two prior approved PRs animating [Scatter/MultiScatter](https://github.com/ihmeuw/ihme-ui/pull/142) and [Line, MultiLine, and Area components](https://github.com/ihmeuw/ihme-ui/pull/146).

In addition to small touchups to those efforts, I have added some more documentation in .src/ui/shape/readme.md.

At one point, Aaron had wanted to include a start on an all encompassing demo app. I ended up scrapping that since this is already quite a bit to bring into one PR. The work has been saved in source control so it should be easy to retrieve if/when we'd like to take that effort on.


jimvermillion:
I've taken time to improve the codecov in the scatter component. Unfortunately, I was unable to address the codecov `miss` (-0.26%) in utils/index.js. The definitions that comprise the `miss` are only react prop types. I do not see this is an issue since there are other prop types within that same directory that are also noted as codecov `miss`es. I'm open to solutions if anyone comes up with one.

Cheers!
-Jim

src/ui/shape/src/area.jsx
```diff
@@ -1,14 +1,21 @@
 import React from 'react';
+import Animate from 'react-move/Animate';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { area } from 'd3';
 import bindAll from 'lodash/bindAll';
+import partial from 'lodash/partial';
```
komalali: Curious as to why the lodash imports are imported separately here but not in demo/app.jsx

src/ui/shape/src/line.jsx
```diff
@@ -1,12 +1,19 @@
 import React from 'react';
+import Animate from 'react-move/Animate';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { line } from 'd3';
 import bindAll from 'lodash/bindAll';
+import partial from 'lodash/partial';
```
komalali: same comment as above. And looking at multi-line.jsx it seems like we're just importing from 'lodash' even when there are only 2 imports.

src/ui/shape/src/area.jsx
jimvermillion: This sub-path pattern was established in geospatial-app to utilize tree-shaking. In general, I think we should work towards it. 

Though `lodash` doesn't have the tree-shaking benefits the way `lodash-es` does, converting to sub-path imports would prime us well for switching to lodash-es. I know waaay less about tree-shaking than @davschne, so he may want to chime in here. That was my overall thinking

As far as why it is not in the demo app, in the course of this work effort, I didn't need to adjust any lodash imports in the demo app. Since this branch has already been quite expansive, I figured it best to stick to it's main purpose as much as possible. I figure a separate `tree-shaking` effort could be done in a subsequent branch. If you think it makes sense to change them in this one, I'm happy to update the demo app's import statements.

davschne: Makes sense to me, Jim. A tree-shaking effort in IHME-UI would be great, but probably best saved for its own PR. The way the library is currently written, it seems to resist all my efforts to tree-shake it in `geospatial-app`. Annoying, because that application only uses a few IHME-UI components.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ihme-ui",
   "description": "Visualization tools from the Institute for Health Metrics and Evaluation",
-  "version": "0.30.1",
+  "version": "0.31.0",
```
davschne: I suggest incrementing the version outside the PR process. When there are multiple open PRs, it's hard to predict the order in which they'll be merged.

src/ui/shape/README.md
```diff
@@ -132,3 +136,130 @@ Property | Required | Type(s) | Defaults | Description
 `shapeType` |  | one of: shapeTypes() | 'circle' | Type of shape to render, driven by d3-shape.<br />One of: 'circle', 'cross', 'diamond', 'square', 'star', 'triangle', 'wye'
 `translateX` |  | number | 0 | Move shape away from origin in x direction.
 `translateY` |  | number | 0 | Move shape away from origin in y direction.
+
```
davschne: Good documentation!

src/ui/shape/src/line.jsx
jimvermillion: *Let me know if you have questions/concerns about the explanation in the above thread.

src/ui/shape/src/area.jsx
davschne: After further reflection and looking at the source code for `geospatial-app`, I realized that we're _not_ using this atomic import pattern for lodash in that repo. `lodash-es` seems to allow tree-shaking when importing the normal way, e.g.
```javascript
import { bindAll, partial } from 'lodash-es';
```
So, if we do make the switch to `lodash-es` here, we should be able to use that pattern and still get the benefits of tree-shaking. We _did_ have to adopt the atomic import pattern to allow tree-shaking of `rxjs`.

jimvermillion: Weird, I had a vivid recollection of seeing/doing that with `lodash-es`, and definitely with `rxjs`...my b, I'll revert to single import statements.

davschne: Well, you could leave the imports as-is. In their current form, you'll get the advantage of tree-shaking, I believe, without actually doing tree-shaking.

davschne: ... but only if _every_ import from `lodash` uses this format.

jimvermillion: Well, had already reverted to using single import statements. I think that'll be easier to update to lodash-es all at once. I've made a new [issue](https://github.com/ihmeuw/ihme-ui/issues/160) to address that, hopefully that'll be cleaner/more comprehensive in the long run.

jimvermillion:
This PR has 1 approval so far (from David). Unless there is an objection, I plan on merging this PR today or tomorrow. 

I'm happy to give more time to anyone that wants to look at the work. Please let me know if there are any questions or concerns.

Thanks, all. I'm very excited about getting this animation bootstrap into the codebase!

Cheers!
-Jim

src/ui/shape/README.md
```diff
@@ -132,3 +136,130 @@ Property | Required | Type(s) | Defaults | Description
...
+---
+
+### Arguments of the animation methods
+It is often necessary to know what data is to be represented during animation in order to affect it 
+accordingly. The signature has been detailed in the @callback type 
+[AnimateMethodCallback](https://github.com/ihmeuw/ihme-ui/blob/react-move/src/utils/props.js#L296). 
+
+Here's is an example of utilizing the arguments in an animation method:
+```
```
Ryshackleton: Really nice work on the documentation here!  I'm not sure if it actually works or not, but have you tried using the javascript language identifier to turn on syntax highlighting in github?  That would make it look even nicer.  https://help.github.com/articles/creating-and-highlighting-code-blocks/

jimvermillion: Great suggestion! 
I've added the `js` tag to the block start and they do, in fact, work on github. I think it'd be great if someone (great first PR) would want to hunt through the `README`s and see if there are other places to provide this highlighting.

PR #159, opened 2018-06-12
STP-196: Remove canvas dependency

jimvermillion:
This PR removes `canvas-prebuilt` and `node-pre-gyp` from IHME-UI's dependencies.

The entirety of our need for these packages has had to do with unit testing. JSDOM utilizes either `canvas` or `canvas-prebuilt` to implement `document.createElement('canvas')`.

`canvas` and `canvas-prebuilt` has been a pain to work into Travis. Sometimes one works, sometimes the other, sometimes neither. I've patched it 3 times now. One measure I took was to involve the `node-pre-gyp` package for the `canvas-prebuilt` package to install correctly.

I've replaced the canvas modules with a simple function called `createElementMock` which uses JSDOM's implementation for `document.createElement` and amends a `getContext` property if the Element `tagLine` parameter === 'canvas'. So it only modifies the behavior for `canvas` Elements.

I think this is the cleanest solution to keep tests running as they were, avoid issues with Travis, and as an added bonus, we go from `19 vulnerabilities (1 low, 18 moderate)` to `10 vulnerabilities (1 low, 9 moderate)` during the npm audit.

Cheers!
-Jim


jimvermillion:
@davschne, thanks for the review. After reading your comment, I realized that the `font` property is also not even needed. So we're down to the `getContext` right now. Should be very easy to add properties if the need arises.

I've updated the return object of `getContext` to not include the `font` property.

Issue #160, opened 2018-06-14
Tree shaking: use lodash-es


jimvermillion:
We should be working towards this library being tree-shakable. To support that effort, we should switch from the `lodash` library to the `lodash-es` library and update the import statements.

PR #171, opened 2018-09-13
Add bar chart to IHME-UI

kbeame:
Apologies for the large pull request.
This pull-request includes the all of the work that Bao did last summer and my updates based on Gabe and Aaron's initial comments on Bao's work. It also includes the update to React 16 necessary to bring it up to date. It also includes my more recent work to make a composition component that is more flexible than how Bao had his set up.
'Stacked' can be anything from a regular bar-chart to a grouped multi bar-chart to a stacked multi bar-chart. Ideally this one component will be able to handle most options one might need, it just depends on how the data is structured.


src/ui/bar/src/bar.jsx
```diff
@@ -0,0 +1,275 @@
...
+  /**
+   * Initial y position of svg element rect.
+   */
+  y: PropTypes.number,
+
+};
+
+Bar.defaultProps = {
+  fill: 'steelblue',
+  foused: false,
```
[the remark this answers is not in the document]
kbeame: Fun fact- it looks like this typo is consistent throughout a decent amount of the code! Thought not all.

src/ui/bar/README.md
```diff
@@ -0,0 +1,101 @@
...
+#### \<MultiBars />
+`import { MultiBars } from 'ihme-ui'`
+
+Property | Required | Type(s) | Defaults | Description
+:---    |:---      |:---     |:---      |:---    
+`barsClassName` |  |  [CommonPropTypes.className](https://github.com/ihmeuw/ihme-ui/blob/master/src/utils/props.js#L11) |  | className applied to `<Bars />`'s outermost wrapping `<g>`  
+`barsStyle` |  | [CommonPropTypes.style](https://github.com/ihmeuw/ihme-ui/blob/master/src/utils/props.js#L11) |  | inline styles applied to `<Bars />`'s outermost wrapping `<g>`.
+`className` |  | object |  | className applied to outermost wrapping `<g>`
+`clipPathId` |  | string |  |  * If a clip path is applied to a container element (e.g., an `<AxisChart />`), clip all children of `<MultiBars />` to that container by passing in the clip path URL id.
+`colorScale` |  | func |  | If provided and `dataAccessors.fill` is undefined, determines the color of bars.
```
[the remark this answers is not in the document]
kbeame: I think this is completed for all the READMEs

src/ui/bar/demo/app.jsx
```diff
@@ -388,45 +388,45 @@ class App extends React.Component {
...
-         selection={this.state.selectedItems}
-         selectedStyle={{
-         stroke: '#000',
-         strokeWidth: 1,
-         }}
-         stacked
+           colorScale={colorScale}
+           data={locationData}
+           dataAccessors={{
+           fill: yearField,
```
Ryshackleton: Seems like 404-408 should also be tabbed in one more step to lint correctly.

src/ui/bar/demo/app.jsx
```diff
@@ -478,47 +478,46 @@ class App extends React.Component {
...
-           selection={this.state.selectedItems}
-           selectedStyle={{
-           stroke: '#000',
-           strokeWidth: 1,
-           }}
-           stacked
+             colorScale={colorScale}
+             data={locationData}
+             dataAccessors={{
+             fill: yearField,
```
Ryshackleton: Same here with one more tab in lines 495-498.

src/ui/bar/demo/app.jsx
```diff
@@ -478,47 +478,46 @@ class App extends React.Component {
...
+             key: 'id', // rename to relate to inner grouping
+             layer: yearField,
+             value: populationField,
+             stack: locationField
+             }}
+             fieldAccessors={{
+             data: 'values',
```
Ryshackleton: Same here.

src/ui/bar/demo/app.jsx
```diff
@@ -478,47 +478,46 @@ class App extends React.Component {
...
+             key: 'key',
+             }}
+             focus={this.state.focus}
+             focusedStyle={{
+             stroke: '#000',
```
Ryshackleton: and here..

src/ui/bar/demo/app.jsx
```diff
@@ -388,45 +388,45 @@ class App extends React.Component {
...
+           layer: yearField,
+           value: populationField,
+           }}
+           fieldAccessors={{
+           data: 'values',
+           key: 'key',
+           }}
+           focus={this.state.focus}
+           focusedStyle={{
+           stroke: '#000',
```
Ryshackleton: Same here.

src/ui/bar/demo/app.jsx
```diff
@@ -0,0 +1,573 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { bindAll, maxBy, minBy, map, uniqBy, xor } from 'lodash';
```
komalali: I've seen us using lodash-es imports in other places to support tree-shaking as mentioned here: https://github.com/ihmeuw/ihme-ui/issues/160

There's some examples in this file: https://github.com/ihmeuw/ihme-ui/blob/master/src/ui/shape/src/shape.jsx

We definitely haven't made a consistent effort to do this but it would be nice in new code?

src/ui/bar/src/bar.jsx
```diff
@@ -0,0 +1,280 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import assign from 'lodash/assign';
```
komalali: Ah yeah, I guess the lodash-es imports are being used in some of these files so it would be good to have it consistent.

kbeame: Ya, it looks like overall in IHME-UI it mostly does not use lodash-es imports but I agree that it would be better for tree-shaking so I'll make the changes in the bar components and stacked composition.

src/ui/bar/demo/app.jsx
```diff
@@ -0,0 +1,579 @@
...
+           xDomain={yearFieldDomain}
+           yDomain={populationFieldDomain}
+           xScaleType="band"
+           yScaleType="linear"
+         >
+         <XAxis />
+         <YAxis />
+         <Bars
+           fill="steelblue"
+           data={data.filter((datum) => { return datum.location === 'India'; })}
```
jimvermillion: Any reason not go short-hand here?
`data={data.filter(datum => datum.location === 'India')}`

src/ui/bar/src/bars.jsx
```diff
@@ -0,0 +1,423 @@
...
+      'onMouseOver',
+      'selectedClassName',
+      'selectedStyle',
+    ]);
+
+    // Given the orientation, set the scales accordingly for the x & y axis
+    const ordinalScale = (isVertical(orientation) ? scales.x : scales.y);
+    const linear = (isVertical(orientation) ? scales.y : scales.x);
+
+    // Check the padding properties and sets it accordingly.
```
davschne: Problematic wording:
- verb mood disagreement ("check" versus "sets")
- To what does "it" refer?

src/ui/bar/src/bars.jsx
```diff
@@ -0,0 +1,423 @@
...
+      bandPaddingOuter
+    );
+
+    return (
+      <g
+        className={className && classNames(className)}
+        clipPath={clipPathId && `url(#${clipPathId})`}
+        style={this.combineStyles(style, data)}
+        transform={`translate(${isVertical(orientation) ? categoryTranslate : 0},
+           ${isVertical(orientation) ? 0 : categoryTranslate})`}
```
davschne: The formatting makes this a little bit challenging to decipher. How about this instead:
```javascript
transform={`translate(
  ${isVertical(orientation) ? categoryTranslate : 0},
  ${isVertical(orientation) ? 0 : categoryTranslate}
)`}
```

src/ui/bar/src/bars.jsx
```diff
@@ -0,0 +1,423 @@
...
+            return (
+              <Bar
+                className={rectClassName}
+                key={key}
+                datum={datum}
+                x={renderingProps.xPosition}
+                y={renderingProps.yPosition}
+                height={renderingProps.barHeight}
+                width={renderingProps.barWidth}
+                fill={colorScale && isFinite(fillValue) ? colorScale(fillValue) : fill}
```
davschne: I'm having a hard time remembering precedence rules for ternary operators versus logical operators. Parentheses in this expression would help clarify things.

src/ui/bar/src/multi-bars.jsx
```diff
@@ -0,0 +1,310 @@
...
+  isVertical,
+  combineStyles,
+  CommonDefaultProps,
+  CommonPropTypes,
+  memoizeByLastCall,
+  propResolver,
+  stackedDataArray,
+} from '../../../utils';
+
+export default class MultiBars extends React.Component {
```
davschne: `React.PureComponent`?

src/ui/bar/src/multi-bars.jsx
```diff
@@ -0,0 +1,310 @@
...
+    // Updates specific domains for each type of bar chart
+    if (stacked) {
+      const stackedDomain =
+        [0, max(plotData, datum => max(datum, d => d[1]))];
+      const linear = (isVertical(orientation) ? scales.y : scales.x);
+      linear.domain(stackedDomain);
+    } else {
+      const ordinal = (isVertical(orientation) ? scales.x : scales.y);
+      layerOrdinal.domain(layerDomain).range([0, ordinal.bandwidth()]);
+    }
```
davschne: I find this tough to reason about, because of the tricky mutation. Is there a way to do this without mutating `scales.x` or `scales.y`?

src/ui/bar/src/multi-bars.jsx
```diff
@@ -0,0 +1,310 @@
...
+    ]);
+
+    return (
+      <g
+        className={className && classNames(className)}
+        clipPath={clipPathId && `url(#${clipPathId})`}
+        style={this.combineStyles(style, data)}
+      >
+        {
+          map(plotData, (datum) => {
```
davschne: I'd extract this anonymous function as a method, `renderBars`. I think it's too important and too complex to be anonymous.

src/ui/bar/test/bars.test.js
```diff
@@ -0,0 +1,238 @@
...
+
+describe('<Bars />', () => {
+  const yearField = 'year_id';
+  const populationField = 'population';
+
+  const data = dataGenerator({
+    primaryKeys: [
+      { name: 'location', values: ['Brazil', 'Russia', 'India', 'China', 'Mexico',
+        'Indonesia', 'Nigeria', 'Vietnam'] }
+    ],
```
davschne: Code formatting could be better here:
```javascript
primaryKeys: [{
  name: 'location',
  values: ['Brazil', 'Russia', 'India', 'China', 'Mexico', 'Indonesia', 'Nigeria', 'Vietnam'],
}],
```

src/ui/bar/test/bars.test.js
```diff
@@ -0,0 +1,238 @@
...
+      selection={[dataFiltered[1], dataFiltered[3]]}
+      selectedClassName="selected"
+      selectedStyle={{ stroke: 'aqua' }}
+      rectClassName="symbol"
+      rectStyle={{ fill: 'bluesteel' }}
+      {...chartDimensions}
+    />
+  );
+
+  it('renders 10 bar', () => {
```
davschne: A test should really describe what the component does universally, not what it does for a very specific input. This component doesn't always render ten bars; it renders ten (I assume) if you pass it a `data` array of length 10.

src/ui/bar/test/bars.test.js
```diff
@@ -0,0 +1,238 @@
...
+    const wrapper = shallow(component);
+    expect(wrapper.find(Bar)).to.have.length(10);
+  });
+
+  it('does not pass specific properties to its children', () => {
```
davschne: I think the test that specifies what props _are_ passed is valuable. This one I find less useful, and it may increase the cost of maintaining the code. I'd consider eliminating it.

src/ui/bar/test/bars.test.js
```diff
@@ -0,0 +1,238 @@
...
+    const assertion = (bar) => {
+      inheritedProps.forEach(prop => {
+        expect(bar).to.have.prop(prop);
+      });
+    };
+
+    shallow(component).find(Bar).forEach(assertion);
+  });
+
+  it('selects a bar', () => {
```
davschne: Selects a bar always, or only when you pass it the appropriate props? Clarification needed.

src/ui/bar/test/bars.test.js
```diff
@@ -0,0 +1,238 @@
...
+    const wrapper = shallow(component);
+    expect(wrapper.find({ selected: true })).to.have.length(2);
+  });
+
+  it('focuses a bar', () => {
```
davschne: Same comment as above.

src/ui/bar/test/bars.test.js
```diff
@@ -0,0 +1,238 @@
...
+    const wrapper = shallow(component);
+    expect(wrapper.find({ focused: true })).to.have.length(1);
+  });
+
+  it('has correct values for the rendering props', () => {
```
davschne: This test probably isn't necessary. This should be covered by the `propTypes` checks.

src/ui/bar/test/bars.test.js
```diff
@@ -0,0 +1,238 @@
...
+      expect(bar).to.have.prop('x').that.is.a('number');
+      expect(bar).to.have.prop('y').that.is.a('number');
+      expect(bar).to.have.prop('height').that.is.a('number');
+      expect(bar).to.have.prop('width').that.is.a('number');
+    };
+    shallow(component).find(Bar).forEach(assertion);
+  });
+
+  describe('selection', () => {
+    it('renders selected bar last', () => {
```
davschne: "last" in what sense?

src/ui/bar/test/bars.test.js
```diff
@@ -0,0 +1,238 @@
...
+        .first()
+        .prop('datum')
+      ).to.equal(selectedDatum);
+      wrapper.setProps({ selection: [selectedDatum] });
+      expect(wrapper
+        .find('g')
+        .find(Bar)
+        .first()
+        .prop('datum')
+      ).to.not.equal(selectedDatum);
```
davschne: This assertion seems possibly unnecessary, because you're already asserting (below) that the selected datum is last.

src/ui/bar/test/bars.test.js
```diff
@@ -0,0 +1,238 @@
...
+      );
+
+      wrapper.find('g').find(Bar).forEach((node, idx) => {
+        expect(node.prop('datum')).to.equal(dataFiltered[idx]);
+      });
+
+      const selectedDatum = dataFiltered[0];
+      wrapper.setProps({ selection: [selectedDatum] });
+      const expectedBarOrder = drop(dataFiltered);
+      expectedBarOrder.push(selectedDatum);
```
davschne: I'm confused about what's going on here.

src/ui/bar/test/multi-bars.test.js
```diff
@@ -0,0 +1,159 @@
...
+  const yearField = 'year_id';
+  const populationField = 'population';
+  const locationField = 'location';
+
+  const data = dataGenerator({
+    primaryKeys: [
+      { name: 'location',
+        values: ['Brazil', 'Russia', 'India', 'China', 'Mexico', 'Indonesia', 'Nigeria',
+          'Vietnam'] }
+    ],
```
davschne: Same comment on code formatting as in `bars.test.js`.

src/ui/bar/test/multi-bars.test.js
```diff
@@ -0,0 +1,159 @@
...
+    { location: 'Indonesia', values: data.filter((datum) => {
+      return datum.location === 'Indonesia';
+    }) },
+    { location: 'Nigeria', values: data.filter((datum) => {
+      return datum.location === 'Nigeria';
+    }) },
+    { location: 'Vietnam', values: data.filter((datum) => {
+      return datum.location === 'Vietnam';
+    }) }
+  ];
```
davschne: Seems like there should be a simpler way of doing this. With this implementation, you have to traverse `data` eight times. How about `lodash.groupBy`?

src/ui/bar/test/multi-bars.test.js
```diff
@@ -0,0 +1,159 @@
...
+    ];
+
+    const assertion = (bars) => {
+      inheritedProps.forEach(prop => {
+        expect(bars).to.have.prop(prop);
+      });
+    };
+
+    shallow(component).find(Bars).forEach(assertion);
+  });
```
davschne: See my comments on `bars.test.js`. Some of them are applicable here as well.

src/ui/compositions/stacked/src/stacked.jsx
```diff
@@ -0,0 +1,367 @@
...
+} from '../../../../utils';
+
+import styles from './style.css';
+import AxisChart from './../../../axis-chart';
+import { XAxis, YAxis } from './../../../axis';
+import MultiBars from './../../../bar/src/multi-bars';
+import ResponsiveContainer from '../../../responsive-container';
+import Legend from './../../../legend';
+
+export default class StackedBarChart extends React.Component {
```
davschne: `React.PureComponent`?

src/ui/compositions/stacked/src/stacked.jsx
```diff
@@ -0,0 +1,367 @@
...
+              colorScale={colorScale}
+              data={data}
+              dataAccessors={dataAccessors}
+              fieldAccessors={fieldAccessors}
+              fill={fill}
+              focus={focus}
+              focusedStyle={{
+                stroke: '#000',
+                strokeWidth: 2,
+              }}
```
davschne: Don't create objects that never change in a React method--that increases performance overhead. I suggest creating it in the module scope. That way the same object will be used every time.
```javascript
// near the top of the file:
const FOCUSED_STYLE = { /* ... */ };
```

src/ui/compositions/stacked/src/stacked.jsx
```diff
@@ -0,0 +1,367 @@
...
+              layerDomain={layerDomain}
+              onClick={onClick}
+              onMouseOver={onMouseOver}
+              onMouseMove={onMouseMove}
+              onMouseLeave={onMouseLeave}
+              style={chartStyle}
+              selection={this.state.selectedItems}
+              orientation={orientation}
+              grouped={(type === 'grouped') ? true : undefined}
+              stacked={(type === 'stacked') ? true : undefined}
```
davschne: Is there some reason not to use a simple boolean value?
```javascript
grouped={type === 'grouped'}
stacked={type === 'stacked'}
```

src/ui/compositions/stacked/src/stacked.jsx
```diff
@@ -0,0 +1,367 @@
...
+    );
+  }
+
+  render() {
+    const { className, displayLegend, style } = this.props;
+    return (
+      <div className={classNames(styles['chart-container'], className)} style={style}>
+        {this.renderTitle()}
+        {this.renderBarChart()}
+        {(displayLegend) ? this.renderLegend() : ''}
```
davschne: For the false case, I'd simply render `null`. Empty string seems less semantic.

src/ui/compositions/stacked/test/stacked.test.js
```diff
@@ -0,0 +1,164 @@
...
+
+  const locationData = [
+    { location: 'Brazil', values: data.filter((datum) => { return datum.location === 'Brazil'; }) },
+    { location: 'Russia', values: data.filter((datum) => { return datum.location === 'Russia'; }) },
+    { location: 'India', values: data.filter((datum) => { return datum.location === 'India'; }) },
+    { location: 'China', values: data.filter((datum) => { return datum.location === 'China'; }) },
+    { location: 'Mexico', values: data.filter((datum) => { return datum.location === 'Mexico'; }) }
+  ];
+
+// Should these be passed or calculated from given dataset within the BarChart component?
```
davschne: Is this comment still useful? Seems like a "TODO".

src/ui/compositions/stacked/test/stacked.test.js
```diff
@@ -0,0 +1,164 @@
...
+  const populationFieldDomain =
+    [minBy(data, populationField)[populationField], maxBy(data, populationField)[populationField]];
+  const yearFieldDomain = map(uniqBy(data, yearField), (obj) => { return (obj[yearField]); });
+  const locationFieldDomain =
+    map(uniqBy(locationData, locationField), (obj) => { return (obj[locationField]); });
+  const colorScale = scaleOrdinal(schemeCategory10);
+
+// create items given the data and the fields specified
```
davschne: Ensure comments are indented the same as the surrounding code.

src/ui/compositions/stacked/test/stacked.test.js
```diff
@@ -0,0 +1,164 @@
...
+
+  it('renders a legend', () => {
+    const wrapper = shallow(component);
+    expect(wrapper.find(Legend)).to.have.length(1);
+  });
+
+  it('renders appropriate chart title', () => {
+    const wrapper = shallow(component);
+    expect(wrapper.find('.title-class'))
+      .to.have.text('Population Between 2000-2009');
```
davschne: Bind a variable for this string somewhere. Having two versions of the same string literal increases the cost of maintaining the code.

src/utils/bar.js
```diff
@@ -0,0 +1,242 @@
...
+ * @param collection : Data collection used to convert into a stacked shape
+ * @param layerField : Represents the subcategory of the stack field
+ * @param valueField : Represents the values typically plotted against
+ * @param stackField : Represents the category in which the data should be plotted against
+ * @param dataField : Represents the field to access the data object within the obj object
+ * @param layerDomain : Represents the domain of the subcategory of the stack field
+ * @returns {object} Returns an object that is the d3 preferred stacked shape
+ */
+export function stackedDataArray(collection, layerField, valueField,
+                                 stackField, dataField, layerDomain) {
```
davschne: Looks like Python code formatting!

src/utils/bar.js
```diff
@@ -0,0 +1,242 @@
...
+    scale.paddingOuter(bandPaddingOuter);
+  } else if (bandPaddingInner) {
+    scale.paddingInner(bandPaddingInner);
+  } else {
+    scale.padding(bandPadding);
+  }
+  if (align) {
+    scale.align(align);
+  }
+  return scale;
```
davschne: Ideally, a function (particularly a utility function) should not mutate its arguments.

src/ui/compositions/stacked/src/stacked.jsx
```diff
@@ -0,0 +1,367 @@
...
+    } = this.props;
+    if (!labelAccessors.title) return null;
+    return (
+      <div className={classNames(styles.title, titleClassName)} style={titleStyle}>
+        {labelAccessors.title}
+      </div>
+    );
+  }
+
+  renderLegend() {
```
jimvermillion: Quick question: some legends allow an item to be disabled (say a year), is that something that can happen in this composition? Or would a user need to create that composition themself at this point? I'd imagine that the case is the latter, but just wanted to check.

src/utils/bar.js
```diff
@@ -0,0 +1,242 @@
...
+import { propResolver } from './objects';
+
+/**
+ * Determines the orientation of the bars relative to the default orientation
+ * of vertical bars.
+ * @param orientation : A string that represents the orientation of the chart
+ * @returns {boolean} : Returns whether the given argument is vertical
+ */
+export function isVertical(orientation) {
+  return (orientation.toLowerCase() === 'vertical'.toLowerCase());
```
jimvermillion: the literal, `'vertical'` is indeed lowercase, I don't think it's necessary to call the method `toLowerCase()`

src/utils/bar.js
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/**
+ * Determines the type of the bars relative to the default orientation
+ * of string "default".
+ * @param type : A string that represents the type of the chart
+ * @returns {boolean} : Returns whether the given argument is default
+ */
+export function isDefault(type) {
+  return (type.toLowerCase() === 'default'.toLowerCase());
```
jimvermillion: same note on `toLowerCase()` for the literal string `'default'`

src/utils/bar.js
```diff
@@ -0,0 +1,242 @@
...
+ * within a stacked bar chart.
+ * @param collection : Data collection used to convert into a stacked shape
+ * @param layerField : Represents the subcategory of the stack field
+ * @param valueField : Represents the values typically plotted against
+ * @param stackField : Represents the category in which the data should be plotted against
+ * @param dataField : Represents the field to access the data object within the obj object
+ * @param layerDomain : Represents the domain of the subcategory of the stack field
+ * @returns {object} Returns an object that is the d3 preferred stacked shape
+ */
+export function stackedDataArray(collection, layerField, valueField,
```
jimvermillion: Tiny ihme-ui format nit: I think most functions with longer lists of args tend to follow a one-arg-per-line convention (except if/when Brett got python-y in places :)
 
I'd suggest:
```
export function stackedDataArray(
  collection,
  layerField,
  valueField,
  stackField,
  dataField,
  layerDomain,
) {
```

src/utils/bar.js
```diff
@@ -0,0 +1,242 @@
...
+                                 stackField, dataField, layerDomain) {
+  const categoricalData = collection.map((data, i) => {
```
jimvermillion: This map function for `categoricalData` is a bit hard to follow, but it does seem like it relies on mutating its `insertObject` via an inner `forEach`. Often, when I see a `forEach` mutating something in an outer scope, `reduce` comes to mind as a cleaner solution.

I'd suggest using a reduce instead to avoid mutation.  Something like:

```
  const categoricalData = collection.map((data, index) => {
    const dataLayers = data[dataField].reduce((accum, datum) => {
      const layerName = datum[layerField];
      const layerValue = datum[valueField];
      return {
        ...accum,
        [layerName]: layerValue,
      };
    }, {});

    return {
      id: index,
      [stackField]: data[stackField],
      ...dataLayers,
    };
  });
``` 

 (you should test this cuz I just sketched it out :)

src/ui/compositions/stacked/src/stacked.jsx
kbeame: You are correct that currently a user would need to create that composition themselves.

src/ui/compositions/stacked/demo/app.jsx
```diff
@@ -0,0 +1,644 @@
...
+              yScale: "band",
+              xScale: "linear",
+              yDomain: locationFieldDomain,
+              xDomain: populationFieldDomain,
+            }}
+            selection={this.state.selectedItems}
+            type={'grouped'}
+          />
+           </code></pre> */}
+          <StackedBarChart
```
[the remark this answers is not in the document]
kbeame: For some reason I thought it wasn't as useful, but I can totally add it back in

src/ui/bar/test/multi-bars.test.js
davschne: Much better!

src/ui/bar/src/multi-bars.jsx
davschne: You can consider my other outstanding PR comments optional, but I would like to see this one implemented, assuming you agree with the premise.

src/ui/bar/src/multi-bars.jsx
davschne: After staring at this for a while, I decided the mutation problem is due to the side-effect-heavy design of `d3` rather than any flaw in this code. I think it makes sense to leave it as it stands.

src/utils/bar.js
davschne: Again, I think this is more a problem with the design of `d3` than anything else. I think we can leave this as-is.

src/ui/compositions/stacked/demo/app.jsx
```diff
@@ -0,0 +1,644 @@
...
+
+    bindAll(this, [
+      'onClick',
+      'onMouseLeave',
+      'onMouseMove',
+      'onMouseOver',
+    ]);
+  }
+
+  onClick(event, datum) {
```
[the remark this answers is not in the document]
davschne: I'm still working to resolve this issue. I've figured out why it's happening--the `datum` objects are different for stacked bar charts than for normal bar charts. Instead of the `datum` objects originally passed to the component, they seem to be two-element arrays that result from reshaping the data with the function `d3.stack` [here](https://github.com/ihmeuw/ihme-ui/blob/make-one-barchart-to-rule-them-all/src/utils/bar.js#L24-L61). Since we're not using the original objects passed, reference equality checks for "focused" bars (those we're hovering over) are failing. The solution I'm pondering is to axe `d3.stack` so that we preserve the original `datum` objects. I'm just not sure what, if anything, to replace it with. Suggestions welcome!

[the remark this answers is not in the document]
komalali: I still get updates from when this PR gets new comments and I just wanna say I miss you guys <3

davschne:
I've refactored the bar-chart components in this PR. The biggest changes are:
- reorganization of components: Instead of `Bars` and `MultiBars` (with stacking/grouping logic spread between the two), we now have `Bars`, `GroupedBars`, and `StackedBars`. `Bars` is only for normal bar charts (i.e. a single bar for each category). `StackedBars` and `GroupedBars` are essentially peers of `Bars`. They're responsible for stacking and grouping logic respectively. The "composition" component `StackedBarChart` has been renamed to `BarChart`, since it can make any of the three types of bar chart.
- simplified component APIs: Each of these components now expects a flat array for prop `data`, each element of which will be passed to a `Bar`. The elements in `data` can be basically any object, as long as it's possible to obtain the category and value (and subcategory, for stacked and grouped bars) using the eponymous `dataAccessors`.

PR #175, opened 2018-12-14
Map 2019

jimvermillion:
This PR updates the Map component to handle subnational features that exist within disputes for GBD 2019.

NOTE: This is a breaking change. Older shapefiles won't render the disputed borders correctly after this takes effect. However, updating the shapefile with the builder and changing client logic to just filter out any unneeded topojson features should be more trivial than past implementations. I'll have an example for how codviz does this up internally soon, but the client consuming the `<Map />` component will only need to filter out locations, rather than switching `admin_id` with `loc_id` to correctly color a disputed feature.

Approach: 
This implementation should be more in line with what Evan has done in EPI and David in LBD. Instead of rendering different dispute layers at every admin level, we'll have just one disputes layer generated by the topojson builder. The mesh functionality within map now handles how the borders should look evaluating if the border should treated as 'disputed-borders', 'non-disputed-borders', or 'selected-non-disputed-borders'.

Next steps:
Evan and I have chosen to punt on handling disputed territory hovering to another work effort since we need to get the basic functionality up and running ASAP for GBD 2019 launch in January.

Cheers!
-Jim


src/ui/compositions/map/src/map.jsx
```diff
@@ -68,6 +66,45 @@ export const filterData = memoizeByLastCall((data, locationIdsOnMap, keyField) =
...
+      }
+
+      if (properties.admins) {
+        admins.addEach(properties.admins);
+      }
+    });
+
+    const locationsExclusive = locations.symmetricDifference(admins);
+    const locationsExclusiveHas = locationsExclusive.has.bind(locationsExclusive);
+    const claimantsHas = claimants.has.bind(claimants);
```
davschne: I'd call this function `isClaimant`.

src/ui/compositions/map/src/map.jsx
```diff
@@ -68,6 +66,45 @@ export const filterData = memoizeByLastCall((data, locationIdsOnMap, keyField) =
  *
  */
 export default class Map extends React.Component {
+  static classifyMesh(geometryKeyField, matches, selected = []) {
+    const claimants = new Set();
+    const locations = new Set();
```
davschne: `nonDisputedLocations`

src/ui/compositions/map/src/map.jsx
```diff
@@ -68,6 +66,49 @@ export const filterData = memoizeByLastCall((data, locationIdsOnMap, keyField) =
...
+      if (locId) {
+        nonDisputedLocations.add(locId);
+      } else {
+        claimants.addEach(properties.claimants);
+        admins.addEach(properties.admins);
+      }
+    });
+
+    // List of nonDisputedLocations that don't exist in both `admins` and `nonDisputedLocations`.
+    // eg: admins=[1, 2], nonDisputedLocations=[2, 3] => xorLocations=[1, 3]
```
davschne: It might be helpful to list the types of locations that are included in `xorLocations`:
- (non-disputed) locations adjacent to the current arc that don't administer any (disputed) locations adjacent to the arc
- locations that administer a disputed location adjacent to the current arc that are not themselves adjacent to the arc

src/ui/compositions/map/src/map.jsx
```diff
@@ -68,6 +66,49 @@ export const filterData = memoizeByLastCall((data, locationIdsOnMap, keyField) =
...
+    const xorLocations = nonDisputedLocations.symmetricDifference(admins);
+    // Convenience function that tests if a value is in `xorLocations`
+    const xorLocationsHas = xorLocations.has.bind(xorLocations);
```
davschne: `isXorLocation`, perhaps, to match the other names here?

jimvermillion: done

src/ui/compositions/map/src/map.jsx
jimvermillion: done

jimvermillion:
@kbeame @evanl5 
:AUCTIONEERS_VOICE: 2 approvals, we've got 2 approvals, do we have 3 approvals? or 4 approvals?

PR #182, opened 2019-04-04
Fix animate Scatter resize issue

kbeame:
NodeGroup will only update when data changes, this means that it will not resize on window change until the data is also changing. To fix this issue I have made it so that we do not hit NodeGroup (renderAnimatedScatter) when the window is resized but will if the data is updated.

### DIFF
--- a/src/ui/shape/src/scatter.jsx
+++ b/src/ui/shape/src/scatter.jsx
@@ -186,12 +186,8 @@ export default class Scatter extends React.PureComponent {
     );
   }
 
-  shouldAnimate() {
-    return (!!this.props.animate && !this.state.windowResize);
-  }
-
   render() {
-    if (this.shouldAnimate()) {
+    if (this.state.shouldAnimate) {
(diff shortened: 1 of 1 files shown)


src/ui/shape/src/scatter.jsx
```diff
@@ -477,4 +474,12 @@ Scatter.propUpdates = {
...
     };
   },
+  shouldAnimate: (state, _, prevProps, nextProps) => {
+    const prevWindowResize = [prevProps.height, prevProps.width];
+    const nextWindowResize = [nextProps.height, nextProps.width];
+    return {
+      ...state,
+      shouldAnimate: (nextProps.animate && isEqual(prevWindowResize, nextWindowResize)),
+    };
+  },
```
davschne: Looks good, Kat! Couple of small suggestions:
- I'd call those variables `prevWindowSize` and `nextWindowSize`.
- The parentheses in the final expression aren't needed.

PR #183, opened 2019-05-09
Big5 v 471 bar chart styling

benbenbuhben:
Still a lot of work that needs to be done (tests, demos, docs, code re-organization). However, this should suffice for a pre-release.

### DIFF
--- a/src/utils/strings.js
+++ b/src/utils/strings.js
@@ -1,5 +1,39 @@
 /* global window */
 
+import memoize from 'lodash/memoize';
+
+const getCanvasContext = memoize(() => {
+  if (!window) {
+    return false;
+  }
+
+  return window.document.createElement('canvas').getContext('2d');
+});
+
+/**
+ * Measure rendered string dimentions
+ * @param {String} str -> the string to measure
+ * @param {String} font -> https://developer.mozilla.org/en-US/docs/Web/CSS/font
+ * @param {CanvasRenderingContext2D} [canvasContext]
+ * @return {object}
+ */
+export const getRenderedStringDimensions = (
+  str = '',
+  font = '10px Helvetica',
+  canvasContext = getCanvasContext(),
+) => {
+  if (!canvasContext && !window) {
+    return 0;
+  }
+
+  const context = canvasContext;
+
+  context.font = font;
+
+  // see https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics
+  return context.measureText(str);
+};
+
 /**
  * Measure rendered width of string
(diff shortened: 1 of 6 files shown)


src/utils/strings.js
```diff
@@ -20,3 +20,16 @@ export const getRenderedStringWidth = (str = '', font = '12px Verdana', canvasCo
   const metrics = context.measureText(str);
   return Math.ceil(metrics.width);
 };
+
+export const sizeOfLongestRotatedString = (
+  values,
+  rotationAngle = -45,
+) => values.reduce((result, label) => {
+  const width = Math.floor(getRenderedStringWidth(label)) + 5;
+  const height = 10;
```
evanl5: It would be nice to calculate height the same way width is calculated

evanl5:
I suggests refactoring  `src/utils/strings.js` into something like this.

```javascript
/* global window */

import memoize from 'lodash/memoize';

const getCanvasContext = memoize(() => {
  if (!window) {
    return false;
  }

  return window.document.createElement('canvas').getContext('2d');
});

/**
 * Measure rendered string dimentions
 * @param {String} str -> the string to measure
 * @param {String} font -> https://developer.mozilla.org/en-US/docs/Web/CSS/font
 * @param {CanvasRenderingContext2D} [canvasContext]
 * @return {object}
 */
export const getRenderedStringDimentions = (str = '', font = '12px Verdana', canvasContext = getCanvasContext()) => {
  if (!canvasContext && !window) {
    return 0;
  }

  context.font = font;

  // see https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics
  return context.measureText(str);
};

/**
 * Measure rendered width of string
 * @param {String} str -> the string to measure
 * @param {String} font -> https://developer.mozilla.org/en-US/docs/Web/CSS/font
 * @param {CanvasRenderingContext2D} [canvasContext]
 * @return {number}
 */
export const getRenderedStringWidth = (str = '', font = '12px Verdana', canvasContext = getCanvasContext()) => {
  const { width } = getRenderedStringDimentions(str, font, canvasContext)
  return Math.ceil(width);
};

/**
 * Measure rendered height of string
 * @param {String} str -> the string to measure
 * @param {String} font -> https://developer.mozilla.org/en-US/docs/Web/CSS/font
 * @param {CanvasRenderingContext2D} [canvasContext]
 * @return {number}
 */
export const getRenderedStringHeight = (str = '', font = '12px Verdana', canvasContext = getCanvasContext()) => {
  const { height } = getRenderedStringDimentions(str, font, canvasContext)
  return Math.ceil(height);
};
```

PR #192, opened 2020-03-17
STP 437 disappearing borders

evanl5:
Fixes issue with country borders disappearing sometimes
Fixes issue with disputed borders not displaying properly


evanl5:
> This all looks fine to me. I didn't dig into the patch you added, but I can if you want. It would help to have a bit of explanation as to what the patch is doing in the PR comments.

The patch enables the mesh filter functions to receive all matched features instead of just the first and last. This is necessary to make the disputed borders display properly. This PR is simply fixing the patch since it appears to have been made for the wrong module and wasn’t being applied because patch-package wasn’t  installed or used.

kbeame:
Looks fine to me except for the issue with the tests not passing (looks like it could be an enzyme version issue?). Also perhaps we should merge the automated pull requests in before re-deploying ihme-ui. Thoughts?

evanl5:
Does anyone know how to add other people to PRs? I can't figure out how to add Melissa.

evanl5:
> Looks fine to me except for the issue with the tests not passing (looks like it could be an enzyme version issue?). Also perhaps we should merge the automated pull requests in before re-deploying ihme-ui. Thoughts?

I was able to fix the tests by using exact versions of react

PR #198, opened 2020-04-15
Update how disputed locations are matched in map.jsx

kbeame:
Update IHME-UI map to allow for the new way that we handle disputed territories.

### DIFF
--- a/src/ui/compositions/map/demo/app.jsx
+++ b/src/ui/compositions/map/demo/app.jsx
@@ -9,6 +9,7 @@ import {
   get as getValue,
   reduce,
   xor,
+  uniqueId,
 } from 'lodash';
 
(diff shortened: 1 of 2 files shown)


src/ui/compositions/map/demo/app.jsx
```diff
@@ -167,6 +209,10 @@ class App extends React.Component {
           focus={this.state.focus}
           geometryKeyField={`properties.${keyField}`}
           colorAccessor={colorAccessor}
+          layerStyle={(layer) => {
+            if (layer === 'disputes') return { fill: 'transparent' };
```
Ryshackleton, line 213: This could be a ternary.

michael-fernandes:
Looks good!

PR #204, opened 2021-02-17
Stp 616 bar chart tickmarks

benbenbuhben:
Made some minor changes to allow for an axisStyle prop in the AxisChart component and also exposed this prop on the bar chart composition.

**_NOTE:_** I was trying to publish a release candidate to npm and accidentally made a minor release from this branch. Everything seems to be working fine and nothing should be affected since it's just the addition of new props.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "ihme-ui",
-  "version": "1.1.7",
+  "version": "1.1.6-rc1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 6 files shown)


pulchrit (approved):
This looks great as I can tell. : )

Ryshackleton (approved):
Looks ok to me. I didn't pull it down and test anything though.

kbeame (approved):
Looks good! It could be nice to include an example of its use in one of the demos but we don't do that with every property so isn't a huge deal.

michael-fernandes (approved):

PR #212, opened 2021-11-18
FGH-681 map accessibility

pulchrit:
This PR makes some accessibility and other improvements to the map zoom and legend controls.

-  I added a classname for the zoom controls button to the Map composition so that it could be passed to Choropleth and the Zoom Controls components. It looks like it was just skipped in the past.
- I changed the styles on the zoom controls to allow for the focus ring to be seen entirely and increased the size of the buttons so that a target area of 48px square (https://hub.ihme.washington.edu/display/DVT/Styling+checklist) was available. I also increased the the font size of the +, - and *. (Focus ring won't be visible in ihme-ui as it has no focus-visible styling enabled.)
- Following the basics of the design pattern for slider (https://www.w3.org/TR/2017/REC-wai-aria-1.1-20171214/#slider), I added keyboard listeners and handlers to allow for keyboard interaction of the choropleth legend slider handles. We use Interact.js for the mouse listeners and handlers, but that package has no keyboard support (https://github.com/taye/interact.js/issues/275). In order to not make a breaking change, I've not switched to a new package. I've just added keyboard handlers in addition to the mouse handlers. 
- I followed the logic for the mouse handler to create  the keyboard handler. 
- I also made a small change to a prop name for clarity; 'which' has been changed to 'whichSliderHandler'.

To test: 
- npm run demo map
- visit: http://localhost:8888//ihme-ui/src/ui/compositions/map/demo/index.html (change port as necessary depending on what MAMP is running on)

Please let me know of changes. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,25 +65,23 @@ import { AxisChart, linspace } from 'ihme-ui';
 4. Create a new ticket branch and check it out.
 `git checkout -b ticket-branch-name`
 5. Make your code changes. 
-6. From the terminal, navigate to whichever component you are changing.
-`cd src/ui/compositions/map`
-7. Run the demo for that component (e.g., map demo). You'll need to re-run this script every time you save additional changes to rebuild the demo. 
+6. Run the demo for that component (e.g., map demo). You'll need to re-run this script every time you save additional changes to rebuild the demo. 
 `npm run demo map`
-8. Open a browser window and enter the path to the demo. For example, http://localhost:8888//ihme-ui/src/ui/compositions/map/demo/index.html (Change the port number to whichever port you're using for local development in MAMP.)
-9. You'll need to refresh the browser every time you rebuild the demo.
-10. Stage changes and commit as necessary.
+7. Open a browser window and enter the path to the demo. For example, http://localhost:8888//ihme-ui/src/ui/compositions/map/demo/index.html (Change the port number to whichever port you're using for local development in MAMP.)
+8. You'll need to refresh the browser every time you rebuild the demo.
+9. Stage changes and commit as necessary.
 ```
 git status // to see changed files`
 git add filename.js // or git add . to add all modified files
 git commit -m "enter your commit message here" // commit your changes locally
 ```
-11. Push up your working branch in order to create a PR.
+10. Push up your working branch in order to create a PR.
 `git push -u origin working-branch-name`
-12. Go to IHME-UI in Github and click on Compare & Pull Request, then click New Pull Request. 
-13. Select your working branch and which branch you're making the request onto (typically main).
-14. Add a description of the changes.
-15. Add other IHME developers as reviewers
-16. Click Create Pull Request.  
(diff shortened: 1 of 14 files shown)


djpatel2 (approved):
Thanks for implementing these changes @pulchrit!

README.md
```diff
@@ -57,7 +57,33 @@ In support of this, `ihme-ui` exposes both a CommonJS (i.e., `var ihmeUI = requi
 import { AxisChart, linspace } from 'ihme-ui';
 //...
 ```
-
+## Local development and posting pull requests
+1. In Github, click on the Code dropdown/button to clone the repo. You can use either https or ssh. 
+2. Open a terminal, navigate to where you have your repos saved, then clone.
+`git clone theUrlFromGithubgoes.here`
```
Ryshackleton, line 63: Somewhere in here you probably need to say to `cd ihme-ui` and `npm i`.

README.md
```diff
@@ -57,7 +57,33 @@ In support of this, `ihme-ui` exposes both a CommonJS (i.e., `var ihmeUI = requi
...
+3. Be sure to [set up a personal token](https://github.blog/2020-12-15-token-authentication-requirements-for-git-operations/); you'll need it to push up your working branch to the remote.
+4. Create a new ticket branch and check it out.
+`git checkout -b ticket-branch-name`
+5. Make your code changes. 
+6. From the terminal, navigate to whichever component you are changing.
```
kbeame, line 68: You actually do not need to navigate to the component in order to run the demo for that particular component, it can be run in the root directory.

src/ui/compositions/choropleth-legend/src/slider.jsx
```diff
@@ -126,13 +189,16 @@ export default class Slider extends React.Component {
         >
         </rect>
         <SliderHandle
-          which={'x1'}
+          ariaLabel="Change the minimum value displayed on the map."
```
Ryshackleton, line 192: Probably a known issue that's un-fixable, but I noticed that I can only change the slider using the arrow keys, even though VoiceOver tells me to use Ctrl+option+arrow key.

Ryshackleton (approved):
Code looks good to me. I had one small comment re the README. I had some issues with the slider, but it sounds like you are already aware and don't have an easy way to fix them.

src/ui/compositions/choropleth-legend/src/slider.jsx
pulchrit: Yes, exactly. I followed the aria design pattern for a horizontal slider (https://www.w3.org/TR/wai-aria-practices/examples/slider/slider-1.html), so I believe the code is set up correctly. Even the example sliders in the design pattern don't work on VoiceOver, so I think the issue is with VoiceOver.

kbeame (approved):

benbenbuhben (approved):

evanl5 (approved):

